PR #29, opened 2018-11-16
Improved the Docker dev workflow slightly.

BenOvermyer:
Reduce the number of steps necessary to work with Write Freely in a Docker environment.

Also, make it clear that this Docker process isn't ready for production.


thebaer:
Looks good. Thanks @BenOvermyer!

PR #32, opened 2018-11-20
Get versioninfo from the git repository

mrvdb:
Sets value of softwareVer during build

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,10 @@
+GITREV=`git describe --tags`
+LDFLAGS=-ldflags="-X 'github.com/writeas/writefreely.softwareVer=$(GITREV)'"
+
 GOCMD=go
-GOINSTALL=$(GOCMD) install
-GOBUILD=$(GOCMD) build
-GOTEST=$(GOCMD) test
+GOINSTALL=$(GOCMD) install $(LDFLAGS)
+GOBUILD=$(GOCMD) build $(LDFLAGS)
+GOTEST=$(GOCMD) test $(LDFLAGS)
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -1,7 +1,10 @@
+GITREV=`git describe --tags`
```
thebaer: Maybe something like this instead, so we can keep `softwareVer` without the **v** included? (It's used in other parts of the app where we don't want the **v**, including in filenames of the archives we release.)

```suggestion
GITREV=`git describe --tags | cut -c 2-`
```

app.go
```diff
@@ -33,10 +33,12 @@ const (
 
 	serverSoftware = "WriteFreely"
 	softwareURL    = "https://writefreely.org"
-
-	softwareVer = "0.3"
 )
 
+// Software version can be set from git env using -ldflags
+var	softwareVer = "v0.3"
```
thebaer: Could you format this file with `gofmt`? And if we remove the **v** from `$(GITREV)` as mentioned before, we can remove it here.

app.go
```diff
@@ -133,7 +135,7 @@ func pageForReq(app *app, r *http.Request) page.StaticPage {
 	p := page.StaticPage{
 		AppCfg:  app.cfg.App,
 		Path:    r.URL.Path,
-		Version: "v" + softwareVer,
```
thebaer: We can leave this as-is if we remove the **v** from `$(GITREV)` as mentioned before.

thebaer:
Looks great, thanks!

mrvdb:
whoohoo!

Issue #5, opened 2018-11-10
Config bind address


koehn:
I'm running writefreely in a Docker container and I'd like it to bind to other than localhost:8080 (preferably 0.0.0.0:8080) so I can forward connections into it from my web server. How can I configure this?

koehn:
never mid; it doesn't actually bind to localhost.

jlelse:
Could you share your Dockerfile?

Fastidious:
@koehn would love to have such dockerfile too. Thanks!

koehn:
You can grab a copy of the Dockerfile I use [from my GitLab snippet](https://gitlab.koehn.com/snippets/83). I publish it on Docker Hub at `koehn/writefreely`. 

If the builders of Writefreely want to include it in their Git repo, they're welcome to do so.

thebaer:
Thanks @koehn! We are working on an official Docker image too: https://phabricator.write.as/T531

mrvdb:
Am I missing something? How can writefreely be bound to 127.0.0.1 only for example when running behind nginx? (was about to open new issue when I saw this one)

mrvdb:
Reopening because I think this is still unresolved, please correct me if I'm missing something here.

Current situation:
- log says 'localhost' while it does not actually bind to localhost
- binding by default to `0.0.0.0` (or ipv6 equiv in dual stack) is perhaps not the best choice

What I personally would like:
1. make the log say what it actually does
2. change the default to actually bind to  `localhost` , or even better `127.0.0.1` / `[::1]`
3. add support for explicit binding (not sure how net/http interprets this, seems to be just an IP string)

PR #38, opened 2018-11-26
Allow 'bind' in config to specify bind address

mrvdb:
Minimal changes, definitely WIP

To resolve:
- how to support dualstack when not using localhost?
- net/http package uses string, mentions IP address instead of bind,
  need info.

For issue #5

### DIFF
--- a/app.go
+++ b/app.go
@@ -404,6 +404,10 @@ func Serve() {
 	http.Handle("/", r)
 
 	// Start web application server
+	var bindAddress = app.cfg.Server.Bind
+	if bindAddress == "" {
+		bindAddress = "localhost"
+	}
 	if app.cfg.IsSecureStandalone() {
 		log.Info("Serving redirects on http://localhost:80")
 		go func() {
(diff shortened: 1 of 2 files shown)


app.go
```diff
@@ -404,6 +404,10 @@ func Serve() {
 	http.Handle("/", r)
 
 	// Start web application server
+	var bindAddress = app.cfg.Server.Bind
+	if bindAddress == "" {
+		bindAddress = "localhost"
+	}
 	if app.cfg.IsSecureStandalone() {
 		log.Info("Serving redirects on http://localhost:80")
```
thebaer: We'll probably want to include the `bindAddress` here too.

thebaer:
Besides the one mentioned change, this looks good. I'm not sure about the net/http stuff off the top of my head, and otherwise for testing I'd like to defer to you or anyone else with a dualstack config. I just don't have the setup to experiment and figure out the best way forward there.

app.go
mrvdb: Yeah, was in doubt about that one. pushed

mrvdb:
I've read up a bit on the methods. The `net/http/Listen` method is basically the same as `net/Listen` with the network set to `tcp`, meaning dualstack tcp.
In practice this means if you use `localhost` the bind will probably go to ipv6 instead of ipv4 (which is correct)
So, we're good to go I think. The only thing to be aware of is the single binding limitation. For example there is no way to bind it to localhost on ipv4 and ipv6 simultaneously. The library is a bit higher level than I thought.

BenOvermyer:
This looks fine to me.

We should really get some integration tests (e.g., InSpec) set up to automatically test variations of this config, but that is probably outside the scope of this PR.

thebaer:
Okay great, I'll merge this now.

PR #39, opened 2018-11-29
Codehighlight, add support for all languages

mrvdb:
This adds support for all highlightjs supported languages.

It determines upfront what js files should be loaded, does that and triggers a highlight on the last one loaded. While we are on the CDN, the common set will still be loaded, so we can be leaner once we go local js and only load the languages we need.

### DIFF
--- a/templates/include/post-render.tmpl
+++ b/templates/include/post-render.tmpl
@@ -22,12 +22,13 @@
     // Given a array of URIs, load them in order
     function loadLanguages(uris, callback) {
       uris.forEach(function(uri) {
-	var sc=document.createElement('script');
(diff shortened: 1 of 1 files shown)


mrvdb:
Anything needed here? From my POV this can be merged. The only ugly thing is the static assignment of the common set, but that will disappear once we go local js and make the default load without any languages.

thebaer:
Nope, this looks great! Nicely done. And I agree on the optimizations with moving to local JS / reducing what languages are loaded by default. That'd be a great thing to tackle next.

Merging now; will include these changes in v0.5.

Issue #42, opened 2018-12-01
Anchor links render proper in Drafts, but not when Published


mrvdb:
Example draft:

  https://qua.name/2m130piy6f

Example post:
  
  https://qua.name/debug/test-for-anchors

Sources are the same

(UPDATE: patch installed on qua.name, so those links don't show the problem anymore)

thebaer:
Yep, this is another hashtag parsing issue. Similar to #6, we shouldn't be replacing "hashtags" inside `href` attributes or the URL part of a Markdown link. It's manifesting this way (working on drafts) because hashtag parsing doesn't run on drafts.

mrvdb:
The situation could be improved with a reasonable change until we come up with something more structural:

1. require hashtags to have spaces around them, thus change the regex to `' #([\p{L}\p{M}\d]+) '` (and adjust replacement code obviously)
2. do hashtag replacement in drafts to so the rendering is the same.

The effect of the first is that the normal hashtags will keep working as intended, the anchors will start working and less code hashtags will be matched. The user has than a bit more control as well.

I've tested the first and that seems to be a reasonable compromise so far for my needs. I haven't checked enabling hashtag replacment for drafts, but the difference caused by this particular issue for anchors is gone anyway

Thoughts?

thebaer:
Added some thoughts on this fix in #43. I'd really like to get a long-term fix in that can address all of these issues, because the current hashtag parsing via regex is really the least-worst option.

mrvdb:
No problem, I'm maintaining local patches anyway for qua.name.  I needed a bit more controle than there is now.

thebaer:
Fixed in #43

Issue #33, opened 2018-11-21
Weird footnote display


dangom:
### Describe the bug
Sometimes the following, in a markdown file:
```
<sup>
    <a id="fnr.1" class="footref" href="#fn.1">1</a>
</sup>
```

Will be transformed into the following html:
```
<sup>
    <a id="fnr.1" class="footref">
        <span>#</span><span class="p-category">fn</span>
    </a>.1”&gt;1
</sup>
```

Some weird interaction of the href hashtag?

### Steps to reproduce (if necessary)
Not sure. In a minimal markdown file things worked ok, so I couldn't reproduce. But checkout broken footnote links [here](https://write.as/dani/writing-a-phd-thesis-with-org-mode).

### Expected behavior

Instead of being displayed like:
<sup><a id="fnr.1" class="footref"><span>#</span><span class="p-category">fn</span></a>.1”&gt;1</sup>

The footnote link would be:
<sup><a id="fnr.1" class="footref" href="#fn.1">1</a></sup>

thebaer:
Fixed in #43

dangom:
This is awesome. When should we expect the changes to land on write.as?!

thebaer:
It's live now! :)

eternaltyro:
> ```
> [1](my-slug#fn.1)
> ```

Is this behavior documented? I'm trying to figure out how to add footnotes.

Issue #6, opened 2018-11-11
"Hashtags" in code blocks get parsed as hashtags


thebaer:
As reported here: https://mst3k.interlinked.me/@cadey/100673777784986480

MCSH:
How about this idea?

instead of using regexp for replacing hashtags, using a parser that does different passes. Like this https://github.com/russross/blackfriday/blob/master/markdown.go#L405

It would be something like this:

assign last_codeblock_end = 0

start = 0
for start < len(doc):
  if start < last_codeblock_end:
    ignore everything
  else:
     check if this is a new code block
     then if not do rendering


I don't think its possible to parse it using regexp, at least not with vanilla regexp, I'm not sure how different Go's regexp is from the regular expression used in theory of computation.

Just for reference, this is the code that needs changing:
https://github.com/writeas/writefreely/blob/master/postrender.go#L33

Hope this helps.

mrvdb:
How can we move forward here? This is probably less trivial than it looks. Some things which stood out to me:

- i would have thought md -> html conversion was a more or less 'solved problem' by now, what makes our conversion special?
- the md parser seems a modified version of https://github.com/russross/blackfriday correct?  some details on why would be helpful
- the extracttags is coming from https://github.com/kylemcc/twitter-text-go which seems to be optimized for tweet parsing. Perhaps not the best choice for extracting hashtags in general?

JLP:
I just am trying out the 0.15.1 version of WriteFreely on paper.wf instance. It looks like the hashtags in code blocks still get treated as hashtags (e.g. if you use #include in C++ code). When I read the post via Mastodon the #include is also added to the list of tags. Should this be reopened or should I open a new issue?

PR #43, opened 2018-12-03
Hashtag linking improvements

mrvdb:
This improves rendering in a number of situations:

- it keeps anchor tags working
- it gives the user some control for not linking, for example in code
  blocks.

Con:
hashTags at the beginning of a line without a space won't get linked.

Workaround related to issues #42 and #6 and #33

### DIFF
--- a/postrender.go
+++ b/postrender.go
@@ -21,7 +21,7 @@ var (
 	endBlockReg     = regexp.MustCompile("</([a-z]+)>\n</(ul|ol|blockquote)>")
 	youtubeReg      = regexp.MustCompile("(https?://www.youtube.com/embed/[a-zA-Z0-9\\-_]+)(\\?[^\t\n\f\r \"']+)?")
 	titleElementReg = regexp.MustCompile("</?h[1-6]>")
-	hashtagReg      = regexp.MustCompile(` #([\p{L}\p{M}\d]+)`)
+	hashtagReg      = regexp.MustCompile(`(?m)(^| )#([\p{L}\p{M}\d]+)`)
 	markeddownReg   = regexp.MustCompile("<p>(.+)</p>")
 )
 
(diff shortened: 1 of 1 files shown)


thebaer:
Thanks for trying to fix this. I agree this would solve some problems (especially for heavy users of same-page anchor tags), but breaking tags at the beginning of a line is too much of a regression to merge this. It also breaks tags that are put together without spaces around them, e.g. only 3 out of 8 tags correctly render in a post like this:

```
#test #lots #of #hashtags
#many#tags#every#where
```

I'm fine with suboptimal fixes sometimes, but especially since:

* it's currently possible to work around the in-page anchor issues, and
* this would introduce a _visible_ bug that fails in these two very common cases,

I'd rather keep the current shortcomings until we have a fix for all scenarios mentioned in those issues.

mrvdb:
Fair enough.  Although, your second example line, is that really a common case? 

The beginning of line failure is indeed a worry. I haven't really tried hard to get that into the regex, which could probably work, using the same method as getting the 'space case' to work.

For the generic case, having a working parser for a non-standardized format, as you mention 'we're not a markdown only shop', will be quite a challenge I think.

mrvdb:
FYI I'm keeping a test page up here for myself which has the changes in this branch: https://qua.name/debug/test-for-hashlink

mrvdb:
No worries. I wanted to have the temporary fixes published somewhere so other instances could choose to patch theirs if they wanted to. I fully agree the permanent fix needs to be different.
Wouldn't it help  if we were coding against a markdown _standard_ like https://commonmark.org or is that a bridge too far?

thebaer:
I don't know that it'd help, since we'll have to extend the standard no matter what. Or is there some other way you were thinking it'd help out?

Either way, I could see the case for adding the _option_ between a standard like commonmark or our restricted flavor of Markdown in the future.

mrvdb:
I was indeed advocating conforming to *a* standard indeed; commonmark seemed the most sensible. 

If that needs to be in the form of an option for compatibility reasons, that is fine obviously

thebaer:
Yeah, I want to support users who just want to write in plain text, and I think the current library is a decent tradeoff between supporting that and basic Markdown without requiring any explicit customization.

But I've added this as a new task: [T557](https://phabricator.write.as/T557)

thebaer:
I'm going to take another crack at the parser-based fix this coming week, but if I'm not able to make progress, I'll merge this in. Since this fixes so many different issues, and hashtags at the beginning of a line work, I think this will be a good solution.

thebaer:
Merging now -- thanks again!

Closes #33, closes #42

PR #44, opened 2018-12-05
Add openssl and root anchors for federation queries

kaiyou:
The lack of trust anchors was preventing outgoing https queries.

The lack of openssl binary was preventing the generation of actor key and cert. In the long term, I think go has proper primitives to access libssl instead of running openssl system calls.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,6 +22,7 @@ RUN mkdir /stage && \
 
 FROM alpine:3.8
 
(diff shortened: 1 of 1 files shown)


thebaer:
I was having a lot of trouble getting Mastodon to accept the keys I was generating with Go's built-in libraries.. Probably need someone with more knowledge in that area to help move off of the openssl system calls. But that would be really great if we could do that.

PR #45, opened 2018-12-08
SQLite support

thebaer:
This adds support for data storage in SQLite, instead of MySQL. [[Phabricator discussion here](https://phabricator.write.as/T529)]

* New optional `[database] type` config value: `sqlite3`
* New config field: `[database] filename` - determines the name of the SQLite database file (default: `writefreely.db`)
* New step in the configuration process that prompts for the database driver and SQLite filename (if that's the chosen option)
* Helper methods in `database.go` for handling differences between MySQL and SQLite
* New `sqlite.sql` file with SQLite-specific queries, bundled with all distributions

These changes also lay the groundwork for future support of other datastores, like PostgreSQL.

Everything worked well in my testing, but I'd like to get more people trying it out and seeing if anything breaks. Also, special thanks to @BenOvermyer for all the up-front work on this!


thebaer:
Running consistently over the weekend and haven't found any other issues. Still considering this "beta" functionality until it's more widely tested, but I think it's good enough to merge now.

PR #54, opened 2019-01-05
Only check for username and password if driver type is mysql

geekgonecrazy:
If you choose sqlite and go through the configure process and then try to start you get this:
```
2019/01/05 20:22:20 Initializing...
2019/01/05 20:22:20 Loading config.ini configuration...
2019/01/05 20:22:20 Loading templates...
2019/01/05 20:22:20 Loading pages...
2019/01/05 20:22:20 Loading user pages...
2019/01/05 20:22:20 Loading encryption keys...
ERROR: 2019/01/05 20:22:20 log.go:26: Database user or password not set.
```

Workaround is to set a username and password even though not needed.  This changes the check to only apply if the driver is mysql

### DIFF
--- a/app.go
+++ b/app.go
@@ -269,7 +269,7 @@ func Serve() {
 		defer shutdown(app)
 
 		schemaFileName := "schema.sql"
-		if app.cfg.Database.Type == "sqlite3" {
+		if app.cfg.Database.Type == driverSQLite {
 			schemaFileName = "sqlite.sql"
 		}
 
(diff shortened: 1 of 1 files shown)


thebaer:
Good catch! Thanks for fixing this.

I made just one change: to use the constants we have set up for the different db drivers, since we should've been doing that already.

Otherwise the patch looks good to me -- we just need to get your signature on the [CLA](https://phabricator.write.as/L1) and then we'll merge it!

thebaer:
No worries -- thanks again!

PR #64, opened 2019-01-09
Respect external $LDFLAGS on build

SamWhited:
I am applying this patch during the Arch build (or will be in the next update) and thought it might be useful for you upstream.

The main feature is that the build will respect external LDFLAGS; this way we can eg. build a position independent executable easier.
It also adds flags by default (they can of course be turned off) to strip the absolute path from debug information so that only the relative path within the repo remains; this way your home directory doesn't end up in the final binary and builds are more reproducible.

I can of course change the defaults to whatever you prefer if you think this patch is useful since they can always be set at build time by changing the appropriate variables.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,10 +1,22 @@
-GITREV=`git describe --tags | cut -c 2-`
-LDFLAGS=-ldflags="-X 'github.com/writeas/writefreely.softwareVer=$(GITREV)'"
+# make(1) from NetBSD (bmake on many Linux's) uses $(.CURDIR), but GNU Make uses
+# $(CURDIR). Normalize on $(.CURDIR) by setting it to $(CURDIR) if not set, and
+# then $(PWD) if still not set.
+.CURDIR ?= $(CURDIR)
+.CURDIR ?= $(PWD)
+
+# For reproducible builds, don't store the current build directory by default;
+# instead store only the path from the root of the repo (so there is still
+# enough info in debug messages to find the correct file).
+GCFLAGS  = -gcflags="all=-trimpath=$(.CURDIR)"
+ASMFLAGS = -asmflags="all=-trimpath=$(.CURDIR)"
+
+GITREV!=git describe --tags | cut -c 2-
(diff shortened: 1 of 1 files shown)


thebaer:
These are really useful changes -- thanks! Removing the absolute path will be a good default to have, particularly for when I make releases.

Just one bit of housekeeping we need and then I'll merge this: could you sign the [CLA](https://phabricator.write.as/L1)?

Also, please feel free to submit other changes that make packaging easier.

SamWhited:
Ah sorry, I don't sign CLAs generally and don't believe they are acceptable in open source software. I seriously doubt this patch is copyrightable, so feel free to use it or copy/paste into your own commit if necessary.

thebaer:
Technically, merely by writing the code you own the copyright to it (at least in the US) :slightly_smiling_face: 

I understand, though. Thanks!

SamWhited:
For non-trivial things sure, but I can't write an if statement and say that no one else can write if statements because I own the copyright. I suspect this is similar because you can't really do it any other way if you were sticking to the stuff you'd already laid out. But of course, I'm not a lawyer and probably don't understand the subtleties of copyright law (which is one of the reasons I don't sign CLAs, I probably have no idea what I'm agreeing too).

thebaer:
Right, but the way in which you _write_ those `if` statements constitutes a larger work, like this pull request. People write poems, articles, and books out of the same words used over and over, yet those inevitably combine to form larger works can be copyrighted as well.

I'm not a lawyer either, but from my layman's understanding, copy and pasting this code wouldn't change anything about who owns the original copyright (you). To commit code I could safely say I own, I'd technically need to write this from scratch (with my original formatting, my original variable names, etc.).

Either way, of course, I wouldn't recommend anyone signing the CLA who doesn't understand what they're agreeing to.

And again, I understand why people might not agree with it. But I do appreciate you bringing up your concerns. It helps influence any potential future decisions we make, e.g. whether or not it's a requirement worth keeping. And thanks again for all the work you've done around packaging this and our other projects!

PR #65, opened 2019-01-16
use font-display:optional to optimize web font loading

nkoehring:
Adds `font-display: optional` to all `@font-face` entries. This optimizes the font loading by allowing the browser to hide the content only for a very short time before either setting the webfont (if loaded by then) or using a default font. It even allows the browser to not load web fonts at all on very slow connections. See the [MDN article about font-display](https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/font-display) for more information.

---

### DIFF
--- a/less/fonts.less
+++ b/less/fonts.less
@@ -3,6 +3,7 @@
 	font-family: 'Open Sans';
 	font-style: normal;
 	font-weight: 400;
+  font-display: optional;
 	src: url('/fonts/open-sans-v13-latin-regular.eot'); /* IE9 Compat Modes */
 	src: local('Open Sans'), local('OpenSans'),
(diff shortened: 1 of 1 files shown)


thebaer:
Thanks for submitting this! One small suggestion: could you make sure the white space is consistent with the rest of the file? Otherwise I just want to see what the new effect looks like -- if it looks good, we'll merge it!

nkoehring:
The effect should be hardly visible. The browser usually waits no longer then 100ms before showing the text (with either the default or the web font, depending on the loading time)

thebaer:
Got it -- after testing and reading through the docs this seems like a reasonable change to make. I wasn't able to see any effect on a simulated slow connection, but we'll keep an eye on it in case it's adversely affecting too many people. Thanks again!

PR #66, opened 2019-01-16
Deploy the app in the proper dir for dependency management

### LINKED ISSUE #57: Docker build doesn't compile static assets
### Describe the bug
v0.7 added a build dependency on go-bindata for including database schema files in the final binary. But when building the Docker image, the static data file isn't generated at the correct time, and the build fails. 

I could use help on this from anyone with more Docker experience, cc @koehn @BenOvermyer @kaiyou @jlelse

1. Run `make build-docker`
1. Wait a while
1. Get `app.go:276:18: undefined: Asset` _(normally defined in the generated `bindata.go` file)_

### Expected behavior
Image successfully builds

### Application configuration
N/A

**Version or last commit**: v0.7.0

kaiyou:
If anyone could please test this, it is meant to close #57. The idea is to copy the source in the right github.com/writeas folder so that dependency management use it instead of re-downloading it from Github, which omits the go-bindata generated file.

---

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,25 +1,24 @@
+# Build image
 FROM golang:1.11.2-alpine3.8 as build
 
 RUN apk add --update nodejs nodejs-npm make g++ git sqlite-dev
-RUN npm install -g less
-RUN npm install -g less-plugin-clean-css
+RUN npm install -g less less-plugin-clean-css
 
-WORKDIR /go/src/app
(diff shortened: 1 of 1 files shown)


thebaer:
Just tested and pushed the latest build to [Docker Hub](https://hub.docker.com/r/writeas/writefreely/tags) -- looks good! Thanks for helping resolve this!

PR #68, opened 2019-01-18
Support user invites

thebaer:
This includes:

- A new `user_invites` config value that determines who can generate invite links
- A new page for generating invite links, with new user navigation link
- A new /invite/ path that allows anyone to sign up via unique invite link, even if registrations are closed
- Tracking who (of registered users) has been invited by whom

It requires an updated database with `writefreely --migrate` in order to work -- be sure to run that before testing these changes.

This closes [T556](https://phabricator.write.as/T556)

### DIFF
--- a/account.go
+++ b/account.go
@@ -45,6 +45,7 @@ type (
 		PageTitle string
 		Separator template.HTML
 		IsAdmin   bool
+		CanInvite bool
 	}
 )
 
@@ -57,6 +58,8 @@ func NewUserPage(app *app, r *http.Request, u *User, title string, flashes []str
 	up.Flashes = flashes
 	up.Path = r.URL.Path
 	up.IsAdmin = u.IsAdmin()
+	up.CanInvite = app.cfg.App.UserInvites != "" &&
+		(up.IsAdmin || app.cfg.App.UserInvites != "admin")
 	return up
 }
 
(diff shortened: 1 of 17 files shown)


mrvdb:
Nice!

For db migrations in general, have you considered using something like https://github.com/golang-migrate/migrate ?

thebaer:
Yep, I looked at that library first -- I was having trouble figuring out how to add migrations for both MySQL and SQLite, and how to set up the files so they'd work with go-bindata. I also looked at https://github.com/pressly/goose, but didn't want people to have to install another binary just to upgrade WF. This probably isn't the best solution for migrations, but I think it's good enough for now (of course, improvements are always welcome).

If you test out these changes, I'd love to hear your feedback!

thebaer:
Not sure what the upgrade steps for Docker users will be with this change. Will it be running something like:

```
docker exec /go/writefreely --migrate
```
?

cc @kaiyou

thebaer:
Will get this merged tomorrow, unless anything comes up!

mrvdb:
On a personal note, not related to the invite feature. 

It would be beneficial if migrations were files (like most of the mentioned libraries seem to support). This would make it more explicit to manage model addition/deviations for implementers by managing them as migrations. At the moment we have one I think (increased fieldsize of post from 16k to 64k)

thebaer:
Yeah, that makes sense, I hadn't thought of that. But I think it'd be a good conversation to have with more instance admins, so we can be sure it works for everyone. And I'm sure we can make a file-based system work with the current implementation down the road -- do you foresee any potential problems with that?

thebaer:
Merging this now, but feel free to continue discussing here or starting a topic on the forums.

thebaer:
Yep, I definitely appreciate it! If you want to start a forum topic on the subject explaining what you'd like to see, please feel free. Otherwise I'll do that when I can.

PR #73, opened 2019-02-01
Changes to app.go to ensure writefreely works with OpenBSD

stevelord:
OpenBSD doesn't seem to want to play with "Asset" in app.go. I've reverted the changes from a previous commit and this works, but I'm not a go developer so I don't know whether this is the right thing to do.

If it is, then great. If not, happy to try things out.

---

### DIFF
--- a/app.go
+++ b/app.go
@@ -15,6 +15,7 @@ import (
 	"flag"
 	"fmt"
 	"html/template"
+	"io/ioutil"
 	"net/http"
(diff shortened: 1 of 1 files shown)


thebaer:
Hey, thanks for taking a look at this! But that error isn't a bug -- just run `make assets` to compile the static assets, and then it should build for you. Let me know if it doesn't, though.

PR #58, opened 2019-01-08
Add go.mod/go.sum

sheenobu:
This could probably be separated out into two separate diffs: 1. go.mod/go.sum and 2. makefile updates. but current master and 0.7.0 is broken with go modules due to go-bindata go get failing (since there is no GOPATH). Feel free to edit/use as needed.

This fixes build issues with:

 * go-bindata, which attempted a go get ...
 * pinning ikeikeikeike/go-sitemap-generator to 1.0.1 instead of master
 * replace all go-get with go build -o tmpbin/

PR #74, opened 2019-02-01
Support Go Modules

SamWhited:
This PR adds support for Go Modules, but does not change anything that could be considered "code" like #58 does. This avoids the previously discussed CLA issues.

The `go get` behavior in modules has been changed to better understand fetching binaries, so I don't think the Makefile needs to change anyways (building with it appears to work fine for me with this patch applied).

---

Because this PR was entirely created with tooling (it would be exactly the same if you ran the tooling and picked the correct versions), I don't believe it is copyrightable and the CLA is not required.


thebaer:
I appreciate the effort, but while you might be right about the copyright issue (I haven't personally verified it), accepting this PR would mean bringing on the additional overhead of maintaining exemptions to the CLA requirement, and potentially setting a precedent for future contributors, which I'd like to avoid.

I went ahead and added support myself -- please feel free to test it out and file reports for any issues you find, and I'll fix them. My own testing around Go modules will be pretty infrequent until I get our dev environments, Write.as build servers, etc. upgraded to Go 1.11.

Thanks for pushing this forward.

PR #81, opened 2019-03-05
Resolve an edge case where last language has error

mrvdb:
If there are multiple language blocks on a page, we set the
onload on the last one to load all highlighting at once.

If the last language block has an error, the onload would
never fire and thus all blocks would not be highlighted.

The simplest resolution is to fire the callback regardless. We've
already loaded everything so running the callback is not causing any
performance hit which is relevant I think.

### DIFF
--- a/templates/include/post-render.tmpl
+++ b/templates/include/post-render.tmpl
@@ -22,8 +22,12 @@
         var sc = document.createElement('script');
         sc.src = uri;
         sc.async = false; // critical?
+	// Set callback on last script
         if (uris.indexOf(uri) == uris.length-1) {
-          sc.onload = callback;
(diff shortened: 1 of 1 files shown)


thebaer:
Thanks -- for my testing, do you have an example post where an error might happen on the last language block?

mrvdb:
[Matt Baer]:
> Thanks -- for my testing, do you have an example post where an error might happen
> on the last language block?

Yes,

https://qua.name/debug/test.md will fail on current master


https://qua.name/debug/test renders properly (patch already applied)

thebaer:
Perfect. Good catch, and thanks for the fix! Merging now.

PR #89, opened 2019-04-04
Support new commentsEnabled property

thebaer:
This is a field previously supported by PeerTube, and just recently added on PixelFed, that should inform other ActivityPub services whether or not comments / replies are enabled on any given post.

WriteFreely doesn't support comments today, so this will always be `false`. When we support them in the future, this value will reflect the author's choice to accept replies or not.

### DIFF
--- a/posts.go
+++ b/posts.go
@@ -1021,7 +1021,10 @@ func fetchPost(app *app, w http.ResponseWriter, r *http.Request) error {
 
 		p.Collection = &CollectionObj{Collection: *coll}
 		po := p.ActivityObject()
-		po.Context = []interface{}{activitystreams.Namespace}
+		po.Context = []interface{}{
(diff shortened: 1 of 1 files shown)


thebaer:
This is most useful when microblogging platforms respect this property, so this PR is waiting on a consensus from maintainers of those projects. So far, it looks like [Pleroma won't support it](https://pleroma.site/objects/6117819a-6b23-437c-b6c7-016c0db2d70f):

> asking software to voluntarily turn off the reply button is not a real solution.

And I'm not sure [what Mastodon maintainers think yet](https://github.com/tootsuite/mastodon/issues/8565#issuecomment-480454880).

thebaer:
Not sure we're going to get buy-in from the Mastodon team either, so I'm going to go ahead and merge.

PR #97, opened 2019-04-30
Tweak Makefile for MacOS

dariusk:
On my MacOS machine, global node binaries like `lessc` are stored in `/usr/local/bin/`. The current Makefile only checks for `/usr/bin/` and this rectifies that error. There is probably a better way to just run `which lessc` and then use that path, defaulting to `node_modules` if that path doesn't exist, but as this is my first contribution to the project I wanted to stick with what was already there, stylistically.

---

### DIFF
--- a/less/Makefile
+++ b/less/Makefile
@@ -1,5 +1,7 @@
 ifeq ($(shell which lessc),/usr/bin/lessc)
 	LESSC=/usr/bin/lessc
+else ifeq ($(shell which lessc),/usr/local/bin/lessc)
+	LESSC=/usr/local/bin/lessc
 else
 	LESSC=node_modules/.bin/lessc
 endif


thebaer:
Hey, thanks for submitting this! These are both great fixes I'd like to get merged. There are just a few potential gotchas with stripping newlines that I want to discuss, so would you mind splitting this into two PRs? That way I can merge the `lessc` Makefile changes while we talk about the newline stuff.

dariusk:
No problem. Will do shortly.

dariusk:
Okay, split this. This PR is now just the `lessc` stuff, #98 is the newlines.

thebaer:
Awesome, thank you! I agree there's probably a better way to use the path from `which lessc` in the future (feel free to improve it anytime), but this will do the trick! Merging now.

PR #98, opened 2019-04-30
Strip newlines from AP Article object HTML output

dariusk:
Some federated services, like Mastodon, interpret newlines in Html literally. Since it costs nothing (and saves a few bytes) to strip newlines from the HTML output, of the Article object, this commit does just that.

What posts look like in Mastodon (modified to have `Article` accepted) before:

image

And after this PR:

image

---

### DIFF
--- a/less/Makefile
+++ b/less/Makefile
@@ -1,5 +1,7 @@
 ifeq ($(shell which lessc),/usr/bin/lessc)
 	LESSC=/usr/bin/lessc
+else ifeq ($(shell which lessc),/usr/local/bin/lessc)
+	LESSC=/usr/local/bin/lessc
 else
 	LESSC=node_modules/.bin/lessc
 endif
--- a/posts.go
+++ b/posts.go
@@ -1057,6 +1057,8 @@ func (p *PublicPost) CanonicalURL() string {
 	return p.Collection.CanonicalURL() + p.Slug.String
 }
 
+var newlineRegex = regexp.MustCompile(`\r?\n`)
+
 func (p *PublicPost) ActivityObject() *activitystreams.Object {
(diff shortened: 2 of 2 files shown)


thebaer:
As I mentioned in #97, I'd definitely like to make this work. But ideally, I'd like to preserve newlines for the cases where they're used for creative effect, as they often are on Write.as / WriteFreely.

One case would be when writing poetry -- someone might use single, not double, line breaks (e.g. [this post](https://write.as/u1v7q02s9axueltv.md)). In those cases, we don't insert `<br />` tags or any other markup indicating the single newline. So ideally, this would only strip newlines between HTML block elements (h1, h2, p, etc.) in the ActivityPub data. And that would have the added bonus of preserving single line breaks in the body text when the data shows up on Mastodon.

dariusk:
That's fair. Anyway, the _actual_ issue here is that Mastodon runs [`simple_format`](https://api.rubyonrails.org/classes/ActionView/Helpers/TextHelper.html#method-i-simple_format) on all incoming text payloads including HTML (this transforms all single-newlines into `<br>`s) so I think I will just fix that instead. This is all for an experimental fork I'm running anyway to support `Article` objects; my hope is to get some of this into upstream but it might be fully rejected.

thebaer:
Ah, gotcha. Seems kind of crazy that they'd run it on HTML as well.

Always happy to consider alternatives on our end, if it'll help.

dariusk:
Oh, they don't actually run it on HTML. It's why they convert everything to a stub! So there's no html at all. I'm working on a branch that accepts a subset of HTML (`strong`, `em`, and other "safe" tags) and am trying to shake out the issues from that.

PR #102, opened 2019-05-10
Decouple writefreely library from executable

thebaer:
This exposes setup and admin functions in the writefreely package, and
uses them in the main application / executable, initialized by the flag parsing that
now happens there.

This is the first step towards making `writefreely` more usable as a
standalone package, and is still a work in progress.


thebaer:
All set for review. Again this is just a ton of refactoring, so we're mostly concerned with making sure nothing broke in the process. Issues are most likely to happen in components that need initialization, like:

* templates
* encryption keys (for session cookies and email addresses stored in database)
* sessions (e.g. not being able to log in)
* the Reader section
* form encoder / decoder (handles API requests)

app.go
```diff
@@ -413,92 +410,178 @@ func Serve() {
 		app.cfg.Database.Database = "writefreely"
 	}
 
+	// TODO: check err
 	connectToDatabase(app)
```
robjloranger: Should we check this error before merging?

app.go
```diff
@@ -413,92 +410,178 @@ func Serve() {
...
+	if _, err := os.Stat(fullKeysDir); os.IsNotExist(err) {
+		err = os.Mkdir(fullKeysDir, 0700)
+		if err != nil {
+			return err
+		}
 	}
+
+	// Generate keys
+	initKeyPaths(app)
+	var keyErrs error
```
robjloranger: Should we use a multierror here? Otherwise we would only report as many as the last of three errors back to the caller.

app.go
thebaer: We'd actually need to refactor `connectToDatabase()` to return an error -- right now it does `os.Exit(1)` when encountering errors, instead. Definitely want to fix it, but I'd say we can save it for a little later.

app.go
thebaer: Possibly.. what's a multierror? I'm not familiar with it.

app.go
robjloranger: :+1:

app.go
robjloranger: It's a hashicorp library, we could use or implement something simliar. not a msut have but it's nice. hashicorp/go-multierror

robjloranger: note: if you go with that don't forget `keys/keys.go`

robjloranger:
:lgtm: :+1:

app.go
thebaer: :+1: thanks, that could definitely be the way to go. I added a TODO in both places so we can be sure to tackle this in another iteration.

thebaer:
Thanks for reviewing! Merging now.

PR #104, opened 2019-05-13
Fix #103: cannot change RTL post direction back to LTR

### LINKED ISSUE #103: Can't change RTL post direction back to LTR
_Originally reported [on discuss.write.as](https://discuss.write.as/t/changing-direction-back-to-left-to-right/599)_:

> Once a post's direction is set to "right-to-left", it can not be changed back to "left-to-right". 
>
> * What happened
> I was playing around with post direction,  toggling right to left option worked, but could not revert it. Page keeps showing "Changes saved" but nothing really happens, the check next to "right to left" just does not go away, and the post direction never changes back.
> 
> * Any steps you took before the bug appeared, if it’s not something obvious
> No. Tried it a couple of times, always the same behaviour. 
> 1. Write a post
> 2. Publish it
> 3. Click Edit next to logo
> 4. Click on "Edit post metadata"
> 5. Toggle direction option. This should work
> 6. Try to toggle it again, never goes away.

cenrak:
[Describe the pull request here...]
Provides a template-based solution to overcome not sending RTL's unchecked box value to the web application.
---

### DIFF
--- a/templates/edit-meta.tmpl
+++ b/templates/edit-meta.tmpl
@@ -255,7 +255,7 @@
 						</select>
 					</dd>
 					<dt><label for="rtl">Direction</label></dt>
-					<dd><input type="checkbox" id="rtl" name="rtl" {{if .Post.IsRTL.Bool}}checked="checked"{{end}} /><label for="rtl"> right-to-left</label></dd>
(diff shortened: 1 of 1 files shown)


thebaer:
Thanks for taking a look at this! However, I think we should go with the API-based solution you mentioned in #103. Ideally, we'd recognize that we're updating post metadata from the web application (instead of the pure API) and set it to `false` if there is no checkbox value.

thebaer:
@cenrak I went with the API-side solution in #110 -- could you check that out and let me know if that works for you?

PR #106, opened 2019-05-17
Add value attribute to RTL checkbox

joyeusenoelle:
As described in the comments of #103.

---

### DIFF
--- a/templates/edit-meta.tmpl
+++ b/templates/edit-meta.tmpl
@@ -255,7 +255,7 @@
 						</select>
 					</dd>
 					<dt><label for="rtl">Direction</label></dt>
-					<dd><input type="checkbox" id="rtl" name="rtl" {{if .Post.IsRTL.Bool}}checked="checked"{{end}} /><label for="rtl"> right-to-left</label></dd>
+					<dd><input type="checkbox" id="rtl" name="rtl" value="true" {{if .Post.IsRTL.Bool}}checked="checked"{{end}} /><label for="rtl"> right-to-left</label></dd>
 					<dt><label for="created">Created</label></dt>
 					<dd>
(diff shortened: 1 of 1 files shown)


templates/edit-meta.tmpl
```diff
@@ -255,7 +255,7 @@
 						</select>
 					</dd>
 					<dt><label for="rtl">Direction</label></dt>
-					<dd><input type="checkbox" id="rtl" name="rtl" {{if .Post.IsRTL.Bool}}checked="checked"{{end}} /><label for="rtl"> right-to-left</label></dd>
+					<dd><input type="checkbox" id="rtl" name="rtl" value="true" {{if .Post.IsRTL.Bool}}checked="checked"{{end}} /><label for="rtl"> right-to-left</label></dd>
```
robjloranger: should this be false and unchecked by default?

joyeusenoelle: The value doesn't need to be `false`; if it's unchecked the element simply won't be included in POST, and we have code on the server side to handle that. The default value should reflect whatever's in the database, which this does. (It's only ever visible when a post has already been stored.)

(There is an argument to be made that we should allow these settings to be changed at the beginning of a post rather than after its first save, but that's not the state of the code at the moment.)

joyeusenoelle: There's also an argument to be made that this should be a select dropdown or a set of radio buttons, especially since the server-side code supports an "auto" setting (that changes RTL status based on the post language) that isn't available here. But I'm not sure if the "auto" code has been written, so I'm leaving that alone for now.

robjloranger: ok, I'm just thinking out loud with your help :smile: 

So if the value doesn't get sent if the box is not checked, how do we change the stored value when unchecking the box and submitting?

robjloranger:
A quick side note, I think this PR should be branched off the`develop` branch and target the same instead of master.

joyeusenoelle:
Oop, probably true.

thebaer:
Hey, and be sure to check out my comments in #104

PR #107, opened 2019-05-18
Add go-bindata as a module dependency

robjloranger:
This adds https://github.com/jteeuwen/go-bindata as a module dependency
as the make target 'build' will result in it's addition.

We could leave it out but user's running `make build` will get unstaged changes as a result. This would only affect those developing writefreely, not so much for end users just building from source.

---

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -32,6 +32,7 @@ require (
 	github.com/guregu/null v3.4.0+incompatible
 	github.com/ikeikeikeike/go-sitemap-generator v1.0.1
 	github.com/imdario/mergo v0.3.7 // indirect
+	github.com/jteeuwen/go-bindata v3.0.7+incompatible // indirect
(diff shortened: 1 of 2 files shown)


thebaer:
Thanks! Could you explain what you mean by "users running `make build` will get unstaged changes as a result"?

robjloranger:
Oh sure, on running that myself from the current develop branch. I then have modifications to the mentioned files as building pulls this in as a dependency. So I see changes in the repo to be staged for a commit.

thebaer:
Ah, got it. So before this PR, running `make build` modifies the go.mod and go.sum files?

robjloranger:
Ya exactly, without already having it installed in your go path. I haven't tested but I assume based on the make file, if you had it installed previously it would not call go get.

thebaer:
Got it, makes sense. Thanks! Merging now.

PR #110, opened 2019-05-19
Enable un-setting RTL setting via web

### LINKED ISSUE #103: Can't change RTL post direction back to LTR

thebaer:
Previously, once RTL was enabled on a post, you couldn't unset it via
the web application. This fixes that. (Fixes #103)

### DIFF
--- a/posts.go
+++ b/posts.go
@@ -67,7 +67,8 @@ type (
 	}
 
 	AuthenticatedPost struct {
-		ID string `json:"id" schema:"id"`
+		ID  string `json:"id" schema:"id"`
+		Web bool   `json:"web" schema:"web"`
 		*SubmittedPost
 	}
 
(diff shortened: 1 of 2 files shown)


thebaer:
Tested via the web and it works. Tested via the API and found no regressions. Merging now.

PR #113, opened 2019-05-28
Modify GetPost() to allow inclusion of pinned posts

### LINKED ISSUE #112: JSON Export: Not exporting pinned posts
### Describe the bug
Currently, when a user exports their data via JSON, posts in a collection which are pinned are not included in the exported JSON file.  This behavior is absent from the txt and csv exports (but IS present for the Prettified export).

1. Create a post.
2. Click on "pin" from that collection's page (i.e. domain-name/my-cool-blog)
3. Go to domain-name/me/export.
4. Click "JSON" or "Prettified" and download the file.
5. Open the JSON file.  The pinned post will be missing from that collection's `posts` value (the array will be empty if there are no unpinned posts in that collection).

### Expected behavior
All posts should be included in the `posts` array, regardless of its `pinned_position` value.

### Application configuration
 - Multi-user mode 
 - Mysql
 - **Open registration?** Yes
 - **Federation enabled?** Yes

**Version or last commit**: 0.9.1

joyeusenoelle:
This PR modifies the `GetPost()` function in [database.go](https://github.com/writeas/writefreely/blob/95e84a1d0e3e23a3156ab624b2f5e9b51ef9a8ba/database.go#L1068-L1123) to add an additional parameter, `includePinned`. When `includePinned` is `false`, the SQL query to fetch posts will include the clause `AND pinned_position IS NULL`; when `includePinned` is `true`, that clause is left out.

In addition, this PR updates all calls to `GetPost()` to include an argument for the new position. Currently this is `false` everywhere but [export.go](https://github.com/writeas/writefreely/blob/95e84a1d0e3e23a3156ab624b2f5e9b51ef9a8ba/export.go#L114), where the argument is `true` to ensure that all posts (including pinned posts) are exported as JSON.

I'm going with this one rather than `fix-json-export` because it's easier to maintain in the future (since we won't have two copies of a largely-identical function to manage).

This should resolve #112.

---

### DIFF
--- a/database.go
+++ b/database.go
@@ -104,7 +104,7 @@ type writestore interface {
 	ClaimPosts(userID int64, collAlias string, posts *[]ClaimPostRequest) (*[]ClaimPostResult, error)
 
 	GetPostsCount(c *CollectionObj, includeFuture bool)
-	GetPosts(c *Collection, page int, includeFuture, forceRecentFirst bool) (*[]PublicPost, error)
+	GetPosts(c *Collection, page int, includeFuture, forceRecentFirst, includePinned bool) (*[]PublicPost, error)
 	GetPostsTagged(c *Collection, tag string, page int, includeFuture bool) (*[]PublicPost, error)
 
 	GetAPFollowers(c *Collection) (*[]RemoteUser, error)
@@ -1065,7 +1065,7 @@ func (db *datastore) GetPostsCount(c *CollectionObj, includeFuture bool) {
 // GetPosts retrieves all standard (non-pinned) posts for the given Collection.
 // It will return future posts if `includeFuture` is true.
 // TODO: change includeFuture to isOwner, since that's how it's used
-func (db *datastore) GetPosts(c *Collection, page int, includeFuture, forceRecentFirst bool) (*[]PublicPost, error) {
+func (db *datastore) GetPosts(c *Collection, page int, includeFuture, forceRecentFirst, includePinned bool) (*[]PublicPost, error) {
 	collID := c.ID
 
 	cf := c.NewFormat()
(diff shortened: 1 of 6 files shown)


thebaer:
Agreed this is the best way to go, and is consistent with how we do this sort of thing in other places. Code looks good at first glance, and I'll test soon.

database.go
```diff
@@ -1089,7 +1089,11 @@ func (db *datastore) GetPosts(c *Collection, page int, includeFuture, forceRecen
 	if !includeFuture {
 		timeCondition = "AND created <= " + db.now()
 	}
-	rows, err := db.Query("SELECT "+postCols+" FROM posts WHERE collection_id = ? AND pinned_position IS NULL "+timeCondition+" ORDER BY created "+order+limitStr, collID)
+	pinnedCondition := ""
+	if !includePinned {
+		pinnedCondition = "AND pinned_position IS NULL"
```
robjloranger: I think you will need a space on the end, `AND pinned_position IS NULL `. Otherwise it will run into `timeCondition`.

joyeusenoelle: The space is included in the query string on line 1096. :)

robjloranger: oh I totally missed that. :blush: 

not that I look though, does white space matter here? I mean extra. for example, if you have only `timeCondition` would the query having two spaces cause an error?

joyeusenoelle: It doesn't. You can see in the deleted row that it's possible to have a double space if `timeCondition` is empty, and SQL doesn't seem to care at all. :)

robjloranger: sweet, thanks for teaching me :)

database.go
```diff
@@ -1065,7 +1065,7 @@ func (db *datastore) GetPostsCount(c *CollectionObj, includeFuture bool) {
 // GetPosts retrieves all standard (non-pinned) posts for the given Collection.
```
thebaer: Just need to update this comment to reflect the option of including pinned posts.

joyeusenoelle: I've updated the docstring as requested. :)

thebaer:
Perfect, thanks!

PR #114, opened 2019-05-30
Include post title in zip export txt files

thebaer:
Previously, when exporting a zip archive, individual .txt files didn't have the post title included. This adds them as Markdown headers to posts that have them.

### DIFF
--- a/export.go
+++ b/export.go
@@ -44,8 +44,9 @@ func exportPostsCSV(u *User, posts *[]PublicPost) []byte {
 }
 
 type exportedTxt struct {
-	Name, Body string
-	Mod        time.Time
+	Name, Title, Body string
+
+	Mod time.Time
 }
 
 func exportPostsZip(u *User, posts *[]PublicPost) []byte {
@@ -67,7 +68,7 @@ func exportPostsZip(u *User, posts *[]PublicPost) []byte {
 			filename += p.Slug.String + "_"
 		}
 		filename += p.ID + ".txt"
-		files = append(files, exportedTxt{filename, p.Content, p.Created})
+		files = append(files, exportedTxt{filename, p.Title.String, p.Content, p.Created})
 	}
 
 	for _, file := range files {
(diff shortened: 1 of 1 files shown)


export.go
```diff
@@ -67,7 +68,7 @@ func exportPostsZip(u *User, posts *[]PublicPost) []byte {
 			filename += p.Slug.String + "_"
 		}
 		filename += p.ID + ".txt"
-		files = append(files, exportedTxt{filename, p.Content, p.Created})
+		files = append(files, exportedTxt{filename, p.Title.String, p.Content, p.Created})
```
robjloranger: just `p.Title`

export.go
```diff
@@ -77,7 +78,12 @@ func exportPostsZip(u *User, posts *[]PublicPost) []byte {
 		if err != nil {
 			log.Error("export zip header: %v", err)
 		}
-		_, err = f.Write([]byte(file.Body))
+		var fullPost string
+		if file.Title != "" {
+			fullPost = "# " + file.Title + "\n\n"
```
robjloranger: if the post has a title, will the header not already be at the top of the post?

robjloranger: oh I see, so somewhere I missed the header is stripped out when creating a txt file from a post?

thebaer: The title is a separate database field, so we need to manually include it like this.

export.go
thebaer: `p.Title` is a `zero.String` (not a `string`), so this get the correct value.

robjloranger: ok, first time I've seen the package

export.go
robjloranger: :+1:

export.go
thebaer: yep, it's just like `sql.NullString` but handles things differently for JSON parsing.

thebaer:
Thanks for reviewing!

Issue #84, opened 2019-03-20
Templates/styling discrepancies


Fastidious:
Using https://blog.lrhale.com/ as an example. If you click/press on "About," see how the font and template itself is different from the one used on the main page? Same when an entry's title is clicked/pressed. Is it possible to keep formatting and styles consistent? If so, how?

thebaer:
There are a few issues here:

* The default styles are different between the blog index and individual blog posts (the About static page and individual posts use the same template)
  * One reason it's done this way is that we want the blog header to display differently between these two templates
* Also, the blog uses a custom stylesheet that doesn't apply the same styles to individual posts and the blog index.

As mentioned in #82, we need to refactor the CSS to be more consistent in how the post body is displayed, because there shouldn't be such a difference, and custom styling should be easier. Thanks for mentioning this

thebaer:
Closed by #115. But let me know if there are still any discrepancies we should address, @Fastidious.

PR #115, opened 2019-06-01
Make post body h2's smaller on index pages

### LINKED ISSUE #82: 2nd (##) and 1st (#) level headings show as same style on blog feed
Hi,

It appears that first and second (maybe more) level headings show as the same size and style when viewing the post from the blog’s index. Clicking on the post then corrects it to the right style. 

Was this intentional?

thebaer:
Previously, \<h2>s in a post were the exact same size as post titles on
index pages (blog index, tag listing). This fixes that by reducing the
font-size of body h2's. Closes #82.

### DIFF
--- a/less/core.less
+++ b/less/core.less
@@ -252,6 +252,8 @@ body {
 				margin-bottom: 0.25em;
 				&+time {
 					display: block;
+					margin-top: 0.25em;
+					margin-bottom: 0.25em;
 				}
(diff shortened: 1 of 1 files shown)


thebaer:
This PR should also close #84.

thebaer:
Thanks, Rob! Merging now.

Issue #96, opened 2019-04-21
API misbehavior with SQLite


WolfgangAxel:
### Describe the bug
In [database.go](https://github.com/writeas/writefreely/blob/e80f99a72edc262ab4ffa65c980313fed8f4ae60/database.go), many SQL calls are made that include `NOW()`, however `NOW()` does not exist in SQLite3 (see [here](https://www.sqlite.org/lang_datefunc.html) and [here](https://www.techonthenet.com/sqlite/functions/now.php) for documentation and examples respectively). As such, endpoints like `/api/me` return errors whenever `access_tokens` are used.

**Edit**: I did not notice before that the `now()` function existed, however, from what I can tell, it is not used in [this line](https://github.com/writeas/writefreely/blob/e80f99a72edc262ab4ffa65c980313fed8f4ae60/database.go#L390), which is causing the issue.

### Steps to reproduce (if necessary)
1. Create a SQLite3 database
2. Create a user account
3. Log in through API (`/api/auth/login`)
4. Use the access token to request user data (`/api/me`)
5. Get `500` error from server (My guess is it is triggered at [this line](https://github.com/writeas/writefreely/blob/e80f99a72edc262ab4ffa65c980313fed8f4ae60/database.go#L420))

### Expected behavior
Should have received user data

### Application configuration
 - **Single mode** 
 - **SQLite3**
 - **Registration Closed**
 - **Federation enabled**

**Version or last commit**: V0.9.1

### Other notes

- The `/api/me/posts` and `/api/me/collections` endpoints return `401` errors, but `/api/me` returns `500`.
- (Unrelated to this bug) `/api/me/posts` and similar should probably support pagination or similar to prevent undue strain on the server / excessive bandwidth usage from users with a large volume of posts.

thebaer:
Thanks for the report! Yes, this query should use the [`now()` helper function](https://github.com/writeas/writefreely/blob/fdbefa806fd05ea5f61cf1a0fc75900c2a1dd24e/database.go#L133) to fix this. Anyone should feel free to patch this -- otherwise I will when I have a chance.

That's also a good point about pagination on the `/api/me/posts` endpoint. We've discussed it a bit for other endpoints [on the forum](https://discuss.write.as/t/api-pagiantion/48) before, but please feel free to start a new topic there for this endpoint in particular.

qwazix:
[...]

PR #119, opened 2019-06-09
Half-fix of #96

qwazix:
This solves the error 500 on the /api/me endpoint but the issue is
not completely solved as this now throws a 401, similar to the other
endpoints, due to SQLite returning an empty result set when queried
with byte array. See the issue discussion for more info.

---

### DIFF
--- a/database.go
+++ b/database.go
@@ -387,7 +387,7 @@ func (db *datastore) GetUserNameFromToken(accessToken string) (string, error) {
 
 	var oneTime bool
 	var username string
-	err := db.QueryRow("SELECT username, one_time FROM accesstokens LEFT JOIN users ON user_id = id WHERE token = ? AND (expires IS NULL OR expires > NOW())", t).Scan(&username, &oneTime)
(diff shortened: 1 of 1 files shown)


thebaer:
Thanks for taking a stab at this. Based on [this Stack Overflow answer](https://stackoverflow.com/a/2336940), I guess `LIKE` compares character by character, where `=` does not, so that's why this works. Great catch.

As for the `db.now()` fixes, does this still work on MySQL? I haven't tested yet, but I _believe_ that having the `db.now()` call as a parameter instead of inline the in query string will mean that it gets converted to text and escaped. E.g. it becomes `expires > "NOW()"` instead of `expires > NOW()`

leo60228:
I'm using the latest commit in this PR and I can login, but all authenticated endpoints fail.

qwazix:
There were a few more places where I should replace `=` with `LIKE`. I still need to check if MySQL expires tokens correctly.

qwazix:
> As for the db.now() fixes, does this still work on MySQL? I haven't tested yet, but I believe that having the db.now()  call as a parameter instead of inline the in query string will mean  that it gets converted to text and escaped. E.g. it becomes expires > "NOW()" instead of expires > NOW()

You were right. I replaced the `?` substitution with string concatenation and tested both MySQL and SQLite with expired tokens and both work as expected.

Pls review.

thebaer:
Tested and verified this works. Thanks!

PR #120, opened 2019-06-11
Support reversing tags

leo60228:
You can now use `/tag:test/rev` instead of `/tag:test/` to reverse the order from the default for that collection.

---

 - I didn't get a confirmation email

### DIFF
--- a/collections.go
+++ b/collections.go
@@ -93,6 +93,7 @@ type (
 	}
 	CollectionFormat struct {
 		Format string
+		Reversed bool
 	}
 
 	collectionReq struct {
@@ -126,6 +127,9 @@ const (
 )
 
 func (cf *CollectionFormat) Ascending() bool {
+	if cf.Reversed {
+		return cf.Format != "novel"
+	}
 	return cf.Format == "novel"
 }
 func (cf *CollectionFormat) ShowDates() bool {
@@ -146,8 +150,8 @@ func (cf *CollectionFormat) Valid() bool {
 }
 
 // NewFormat creates a new CollectionFormat object from the Collection.
-func (c *Collection) NewFormat() *CollectionFormat {
-	cf := &CollectionFormat{Format: c.Format}
+func (c *Collection) NewFormat(reversed bool) *CollectionFormat {
+	cf := &CollectionFormat{Format: c.Format, Reversed: reversed}
 
 	// Fill in default format
 	if cf.Format == "" {
@@ -659,13 +663,13 @@ func checkUserForCollection(app *app, cr *collectionReq, r *http.Request, isPost
 	return u, nil
 }
 
-func newDisplayCollection(c *Collection, cr *collectionReq, page int) *DisplayCollection {
+func newDisplayCollection(c *Collection, cr *collectionReq, page int, reversed bool) *DisplayCollection {
 	coll := &DisplayCollection{
 		CollectionObj: &CollectionObj{Collection: *c},
 		CurrentPage:   page,
 		Prefix:        cr.prefix,
 		IsTopLevel:    isSingleUser,
-		Format:        c.NewFormat(),
+		Format:        c.NewFormat(reversed),
 	}
 	c.db.GetPostsCount(coll.CollectionObj, cr.isCollOwner)
 	return coll
@@ -712,7 +716,7 @@ func handleViewCollection(app *app, w http.ResponseWriter, r *http.Request) erro
 
 	// Fetch extra data about the Collection
 	// TODO: refactor out this logic, shared in collection.go:fetchCollection()
-	coll := newDisplayCollection(c, cr, page)
+	coll := newDisplayCollection(c, cr, page, strings.HasSuffix(r.URL.Path, "/rev") || strings.HasSuffix(r.URL.Path, "/rev"))
 
 	coll.TotalPages = int(math.Ceil(float64(coll.TotalPosts) / float64(coll.Format.PostsPerPage())))
 	if coll.TotalPages > 0 && page > coll.TotalPages {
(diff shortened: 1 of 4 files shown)


mrvdb:
Nice idea! 
Could this also be implemented similar to something like this:

`/tag:test` or `tag:+test` : show collection having tag `test` 
`tag:-test` : show collection **not** having tag `test`

leo60228:
@mrvdb This is the first time I've ever touched Go; I'm not really sure how to implement something like that. What's your usecase for this?

mrvdb:
My apologies, I see I've mis-assumed what this was about. I thought it was about filtering instead of ordering.

For the filtering I had something like this in mind for a while but I have no specific usecase other than the need to filter a collection. Using the `+` and `-`  has a number of advantages in my view:

- it's easy to extend (like `+this+that-somethingelse`) 
- not english language dependent 

Having the sort order reversed is something else indeed. :-)

thebaer:
Thanks for working on this, @leo60228! We have some other things on our plate at the moment, but I do think this would be a great addition to the platform.

My only feedback right now is that I'd prefer to use `/chrono` instead of `/rev`, since:

- It more clearly explains which order posts will be in (chronological)
- It's what Tumblr uses -- so if someone comes from that platform and tries it with WF, it'll just work

Thoughts?

(Also, I like the filtering idea, @mrvdb -- it's been discussed a bit [on the forum](https://discuss.write.as/t/multiple-tags/174). Would love to get your input over there.)

collections.go
```diff
@@ -712,7 +716,7 @@ func handleViewCollection(app *app, w http.ResponseWriter, r *http.Request) erro
 
 	// Fetch extra data about the Collection
 	// TODO: refactor out this logic, shared in collection.go:fetchCollection()
-	coll := newDisplayCollection(c, cr, page)
+	coll := newDisplayCollection(c, cr, page, strings.HasSuffix(r.URL.Path, "/rev") || strings.HasSuffix(r.URL.Path, "/rev"))
```
thebaer: We can remove the duplicate condition here. Also, it should check for `/chrono`

collections.go
```diff
@@ -810,9 +814,10 @@ func handleViewCollectionTag(app *app, w http.ResponseWriter, r *http.Request) e
 		return err
 	}
 
-	coll := newDisplayCollection(c, cr, page)
+	rev := strings.HasSuffix(r.URL.Path, "/rev") || strings.HasSuffix(r.URL.Path, "/rev")
```
thebaer: We can remove the duplicate condition here. Also check for `/chrono` instead.

collections.go
thebaer: We can remove the duplicate condition here. Also should check for `/chrono` instead.

database.go
```diff
@@ -1127,10 +1127,10 @@ func (db *datastore) GetPosts(c *Collection, page int, includeFuture, forceRecen
 // given tag.
 // It will return future posts if `includeFuture` is true.
 // TODO: change includeFuture to isOwner, since that's how it's used
-func (db *datastore) GetPostsTagged(c *Collection, tag string, page int, includeFuture bool) (*[]PublicPost, error) {
+func (db *datastore) GetPostsTagged(c *Collection, tag string, page int, includeFuture bool, reversed bool) (*[]PublicPost, error) {
```
thebaer: Let's use `chrono` or `chronological` as the parameter name, to avoid confusion.

database.go
```diff
@@ -105,7 +105,7 @@ type writestore interface {
 
 	GetPostsCount(c *CollectionObj, includeFuture bool)
 	GetPosts(c *Collection, page int, includeFuture, forceRecentFirst, includePinned bool) (*[]PublicPost, error)
-	GetPostsTagged(c *Collection, tag string, page int, includeFuture bool) (*[]PublicPost, error)
+	GetPostsTagged(c *Collection, tag string, page int, includeFuture bool, reversed bool) (*[]PublicPost, error)
```
thebaer: Let's use `chrono` or `chronological` as the parameter name, to avoid confusion.

routes.go
```diff
@@ -170,6 +170,7 @@ func initRoutes(handler *Handler, r *mux.Router, cfg *config.Config, db *datasto
 func RouteCollections(handler *Handler, r *mux.Router) {
 	r.HandleFunc("/page/{page:[0-9]+}", handler.Web(handleViewCollection, UserLevelOptional))
 	r.HandleFunc("/tag:{tag}", handler.Web(handleViewCollectionTag, UserLevelOptional))
+	r.HandleFunc("/tag:{tag}/rev", handler.Web(handleViewCollectionTag, UserLevelOptional))
```
thebaer: Let's go with `/tag:{tag}/chrono` for the route, instead of `/tag:{tag}/rev`.

collections.go
```diff
@@ -93,6 +93,7 @@ type (
 	}
 	CollectionFormat struct {
 		Format string
+		Reversed bool
```
thebaer: Let's name this `Chronological`, instead, to make this less confusing.

collections.go
```diff
@@ -146,8 +150,8 @@ func (cf *CollectionFormat) Valid() bool {
 }
 
 // NewFormat creates a new CollectionFormat object from the Collection.
-func (c *Collection) NewFormat() *CollectionFormat {
-	cf := &CollectionFormat{Format: c.Format}
+func (c *Collection) NewFormat(reversed bool) *CollectionFormat {
```
thebaer: Let's use `chrono` or maybe `forceChrono` as the parameter name, instead, to make this less confusing.

collections.go
```diff
@@ -126,6 +127,9 @@ const (
 )
 
 func (cf *CollectionFormat) Ascending() bool {
+	if cf.Reversed {
+		return cf.Format != "novel"
```
thebaer: It would be helpful to have some comments here explaining the logic a bit (or some explanation here on github). I'm not sure why we're checking for the `novel` format here instead of always returning `true`.

collections.go
```diff
@@ -659,13 +663,13 @@ func checkUserForCollection(app *app, cr *collectionReq, r *http.Request, isPost
 	return u, nil
 }
 
-func newDisplayCollection(c *Collection, cr *collectionReq, page int) *DisplayCollection {
+func newDisplayCollection(c *Collection, cr *collectionReq, page int, reversed bool) *DisplayCollection {
```
thebaer: Again let's use `chrono` or maybe `forceChrono` as the parameter name, instead, to make this less confusing.

leo60228:
The reason it's `rev` instead of `chrono` is because it uses the opposite of the blog's default ordering, not forcing chronological. For novel, it's in most recent order instead of the default of chronological.

PR #122, opened 2019-06-14
Update script

robjloranger:
add a basic update script for admins to check for and upgrade to the latest version of writefreely

---

### DIFF
--- a/scripts/upgrade-server.sh
+++ b/scripts/upgrade-server.sh
@@ -1,23 +1,23 @@
 #! /bin/bash
 ###############################################################################
-##											writefreely update script														 ##
-##																																					 ##
-##		WARNING: running this script will overwrite any modifed assets or 		 ##
-##		template files. If you have any custom changes to these files you 		 ##
-##		should back them up FIRST.																						 ##
-##																																					 ##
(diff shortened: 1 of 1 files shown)


thebaer:
Thanks @robjloranger! I just made a few minor tweaks to the wording and error handling. Also, I noticed the "current version" output gets messed up when using a custom build, where the version might be something like `0.9.1-50-gac7d727`

I don't think it's crucial to fix this right now for our purposes. But if it's quick to do, we should. What do you think, easy fix?

robjloranger:
> Thanks @robjloranger! I just made a few minor tweaks to the wording and error handling. Also, I noticed the "current version" output gets messed up when using a custom build, where the version might be something like `0.9.1-50-gac7d727`
> 
> I don't think it's crucial to fix this right now for our purposes. But if it's quick to do, we should. What do you think, easy fix?

That should be an easy fix.

robjloranger:
weird it still looks wrong, the spacing in the header

robjloranger:
it looks fine in real life though
Screenshot from 2019-06-14 17-31-31

thebaer:
Awesome, thanks! Spacing looks good :)

PR #123, opened 2019-06-17
Private instances

thebaer:
This changes the `private` config setting to make an instance entirely private. That is, when enabled, blogs will only be readable by other authenticated users on the same instance.

It still needs a bit of testing, so it's a draft for now.

When finished, this closes [T576](https://phabricator.write.as/T576).


posts.go
```diff
@@ -1193,6 +1193,16 @@ func getRawCollectionPost(app *App, slug, collAlias string) *RawPost {
...
+	slug := vars["slug"]
+
+	isJSON := strings.HasSuffix(slug, ".json")
+	isXML := strings.HasSuffix(slug, ".xml")
+	isMarkdown := strings.HasSuffix(slug, ".md")
+	return strings.HasSuffix(slug, ".txt") || isJSON || isXML || isMarkdown
+}
+
 func viewCollectionPost(app *App, w http.ResponseWriter, r *http.Request) error {
 	vars := mux.Vars(r)
```
robjloranger: these 7 lines are redundant now with `isRaw` above.

posts.go
```diff
@@ -1202,12 +1212,6 @@ func viewCollectionPost(app *App, w http.ResponseWriter, r *http.Request) error
 	isMarkdown := strings.HasSuffix(slug, ".md")
 	isRaw := strings.HasSuffix(slug, ".txt") || isJSON || isXML || isMarkdown
 
-	if strings.Contains(r.URL.Path, ".") && !isRaw {
```
robjloranger: should this still check if `isRaw`?

handle.go
```diff
@@ -331,21 +386,21 @@ func (h *Handler) WebErrors(f handlerFunc, ul UserLevel) http.HandlerFunc {
 
 			var session *sessions.Session
 			var err error
-			if ul != UserLevelNone {
+			if ul(h.app.App().cfg) != UserLevelNoneType {
 				session, err = h.sessionStore.Get(r, cookieName)
-				if err != nil && (ul == UserLevelNoneRequired || ul == UserLevelUser) {
+				if err != nil && (ul(h.app.App().cfg) == UserLevelNoneRequiredType || ul(h.app.App().cfg) == UserLevelUserType) {
 					// Cookie is required, but we can ignore this error
 					log.Error("Handler: Unable to get session (for user permission %d); ignoring: %v", ul, err)
```
robjloranger: need to call `ul()` here

handle.go
```diff
@@ -404,21 +475,21 @@ func (h *Handler) Web(f handlerFunc, ul UserLevel) http.HandlerFunc {
 				log.Info("\"%s %s\" %d %s \"%s\"", r.Method, r.RequestURI, status, time.Since(start), r.UserAgent())
 			}()
 
-			if ul != UserLevelNone {
+			if ul(h.app.App().cfg) != UserLevelNoneType {
 				session, err := h.sessionStore.Get(r, cookieName)
-				if err != nil && (ul == UserLevelNoneRequired || ul == UserLevelUser) {
+				if err != nil && (ul(h.app.App().cfg) == UserLevelNoneRequiredType || ul(h.app.App().cfg) == UserLevelUserType) {
 					// Cookie is required, but we can ignore this error
 					log.Error("Handler: Unable to get session (for user permission %d); ignoring: %v", ul, err)
```
robjloranger: again here, call `ul()`

handle.go
```diff
@@ -523,27 +652,27 @@ func (h *Handler) Download(f dataHandlerFunc, ul UserLevel) http.HandlerFunc {
...
 			start := time.Now()
 
 			var status int
-			if ul != UserLevelNone {
+			if ul(h.app.App().cfg) != UserLevelNoneType {
 				session, err := h.sessionStore.Get(r, cookieName)
-				if err != nil && (ul == UserLevelNoneRequired || ul == UserLevelUser) {
+				if err != nil && (ul(h.app.App().cfg) == UserLevelNoneRequiredType || ul(h.app.App().cfg) == UserLevelUserType) {
 					// Cookie is required, but we can ignore this error
 					log.Error("Handler: Unable to get session (for user permission %d); ignoring: %v", ul, err)
```
robjloranger: again here, `ul()`

handle.go
```diff
@@ -23,24 +23,52 @@ import (
...
-	UserLevelOptional                      // user or not -- object fetched if user
-	UserLevelNoneRequired                  // non-user (required)
-	UserLevelUser                          // user (required)
+	UserLevelNoneType         UserLevel = iota // user or not -- ignored
+	UserLevelOptionalType                      // user or not -- object fetched if user
+	UserLevelNoneRequiredType                  // non-user (required)
+	UserLevelUserType                          // user (required)
 )
 
+func UserLevelNone(cfg *config.Config) UserLevel {
```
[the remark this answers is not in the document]
thebaer: Yeah, to satisfy the `UserLevelFunc` type from line 71. The `UserLevelReader()` func does use the config

posts.go
thebaer: The condition will always be false here, as it's now handled in `CollectionPostOrStatic()` in `handle.go`. The move to an additional func was needed to always allow access to static assets, while restricting access to posts if necessary.

posts.go
thebaer: I know it's not the cleanest, but we need the result of `isRaw()` for `CollectionPostOrStatic()`, and we need all the vars in that logic below. I put them close together so we could have both without a ton of refactoring. Maybe for now I should add some comments mentioning that the two places need to be kept in parity?

handle.go
thebaer: Good catch -- will fix these.

thebaer: Done.

posts.go
robjloranger: sure that's probably good

robjloranger:
testing looks good running locally. I haven't tested federation though, did you want me to do that too?

posts.go
thebaer: done :+1:

thebaer:
Thanks @robjloranger -- no need to test federation. Merging now!

PR #127, opened 2019-06-21
Shorter config process

qwazix:
This resolves T657: shorter-config-process by adding a --sections parameter that takes a space separated string of sections (app, db and server are the current valid ones).


qwazix:
Btw, currently the program just ignores `--sections` if `--config` is not present. Do we want to throw an error?

robjloranger:
That might be a good idea, error with the usage output.

qwazix:
Ok will do.

qwazix:
Now that I look at it...

As the code is now the whole flag parsing process is to match the first flag found and then ignore anything else, so it seems a bit out of place to throw an error in this particular combination (or lack thereof).

It's inconsistent to not throw any error at `writefreely --config --gen-keys` and do so at `writefreely --gen-keys --sections app` even though both exhibit the same behavior: they ignore the second flag.

joyeusenoelle:
One approach might be to use `flag.Visit` to count the number of supplied flags; if it's >1 and the flags aren't `--config` and `--sections`, print an error ("One at a time, please") and then continue with the first flag as normal.

qwazix:
I like this, since it will not fail silently in either case, without having to rework the whole thing. Let me take a stab at it.

thebaer:
For some reason my earlier response didn't send, but I'd say yeah, it should fail silently as it does with other flags. The command-line side of WF is used infrequently enough that I'd say it's not worth getting too far in the weeds with it.

qwazix:
Ok then, feel free to merge the PR as is.

thebaer:
Went ahead and fixed these up. Thanks @qwazix! Merging now.

qwazix:
Thanks, I missed that `go fmt`, though I admit I did read about it in the style guide :-/ Will make sure to run it next time.

thebaer:
No worries!

PR #128, opened 2019-06-25
Fix #124 according to the snippet by @mrvdb

### LINKED ISSUE #124: C syntax highlighting doesn't work
### Describe the bug
We're using a stripped-down version of the highlight.js library that doesn't include the needed code to do C syntax highlighting, and the right library isn't being dynamically loaded.

Notice that there is no `c.min.js` file in `static/js/highlightjs/`, so a WriteFreely post [like this](https://pencil.writefree.ly/hello/c-test) attempts to load it, but fails.

### Solution

Instead of loading `c.min.js` when we encounter a code block labeled as `c`, we should load `cpp.min.js`. It lists `c` as an alias (among others, which we should probably do the same for):

```
aliases: ['c', 'cc', 'h', 'c++', 'h++', 'hpp', 'hh', 'hxx', 'cxx'],
```

([from cpp.js](https://github.com/highlightjs/highlight.js/blob/5ab64fc85d74961539f18d0f2061afd4060c3e2a/src/languages/cpp.js#L98))

### Application configuration
 - **Single mode or Multi-user mode?** either
 - **Database?** any
 - **Open registration?** any
 - **Federation enabled?** any

**Version or last commit**: v0.9.1

qwazix:
I changed the sh alias to shell instead of bash.

The additions to the `highlight(nodes)` function look redundant.
It works for me without them but maybe they cover an edge case I
cannot think about?

### DIFF
--- a/app.go
+++ b/app.go
@@ -449,19 +449,21 @@ func DoConfig(app *App) {
 		log.Error("Unable to configure: %v", err)
 		os.Exit(1)
 	}
-	if d.User != nil {
-		app.cfg = d.Config
-		connectToDatabase(app)
-		defer shutdown(app)
+	app.cfg = d.Config
+	connectToDatabase(app)
+	defer shutdown(app)
 
-		if !app.db.DatabaseInitialized() {
-			err = adminInitDatabase(app)
-			if err != nil {
-				log.Error(err.Error())
-				os.Exit(1)
-			}
+	if !app.db.DatabaseInitialized() {
+		err = adminInitDatabase(app)
+		if err != nil {
+			log.Error(err.Error())
+			os.Exit(1)
 		}
+	} else {
+		log.Info("Database already initialized.")
+	}
 
+	if d.User != nil {
 		u := &User{
 			Username:   d.User.Username,
 			HashedPass: d.User.HashedPass,
--- a/database.go
+++ b/database.go
@@ -388,7 +388,7 @@ func (db *datastore) GetUserNameFromToken(accessToken string) (string, error) {
 
 	var oneTime bool
 	var username string
-	err := db.QueryRow("SELECT username, one_time FROM accesstokens LEFT JOIN users ON user_id = id WHERE token = ? AND (expires IS NULL OR expires > NOW())", t).Scan(&username, &oneTime)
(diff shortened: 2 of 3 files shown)


mrvdb:
Notes for consideration:
- `sh` is an alias of `bash` in highlightjs itself, perhaps not deviate from their maps (see : https://github.com/highlightjs/highlight.js/blob/master/src/languages/bash.js#L40 ) I think they interpret `shell` as a console session which has a different begin pattern than `sh`, but otherwise just uses `bash` highlighting.
- the `registerlanguage` parts are to make sure the used language is registered properly (see highlightjs documentation on 'usage') as we load 'on demand' What it does internally, I'm not sure.

qwazix:
* Ok I will revert `sh` to `bash`
* I assume it loads it in the list of supported languages for auto-detection. Since we don't seem to do auto-detection (correct me if I'm wrong) it is probably redundant. I will try to see what it does but highlighting specific code blocks works without it. (There was no registerLanguage before anyway).

qwazix:
Regarding `registerLanguage` it doesn't have anything to do with auto-detection. In the documentation it is referenced only under CommonJS and not for inclusion in the browser. In the following snippet both auto-detection and specific block highlighting works so I'm inclined to keep the `registerLanguage` part out unless there's a demonstrable lack of functionality.

``` javascript
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <meta name="viewport" content="width=device-width, initial-scale=1.0">
    <title>Document</title>
    <link rel="stylesheet" href="//cdnjs.cloudflare.com/ajax/libs/highlight.js/9.15.8/styles/default.min.css">
    <script src="highlight/highlight.pack.js" ></script>
    <script src="highlight/javascript.min.js" ></script>
</head>
<body>
    <pre><code>
        document.querySelectorAll("body.dark");
    </code></pre>
    <!-- <script>hljs.initHighlightingOnLoad();</script> -->
    <script>hljs.highlightBlock(document.querySelector("code"));</script>
</body>
</html>
```

mrvdb:
> I'm inclined to keep the registerLanguage part out 

Agreed. I can't see any useful effect either, I assumed it at least kept the list of languages, but it does not.

qwazix:
Nice catch. Removed it.

thebaer:
Thanks, @qwazix and @mrvdb. This looks great! Merging now.

Issue #125, opened 2019-06-20
Local git tags interfere with versioning


mrvdb:
The build system uses `git describe --tags | cut -c 2-` to construct a version number from tags like `v0.9.1` to display something like `0.9.1-102-g31fa4b0`

The `--tags` option makes it interfere with tags local to the repository which are just pointers and not annotated with a message like the version tags.

Options to resolve the issue:
- omit `--tags` to only consider annotated tags, sufficient to resolve
- use a pattern check like `v*` or something more fancy to filter tags
- use `--contains` option? not sure how that would work

For now, just omitting `--tags` should be fine.

mrvdb:
any comments on this? I can make a PR, but the proposed change is fairly trivial

thebaer:
Yeah, removing `--tags` is a simple enough fix. Please feel free to make a PR, and we'll get it in.

PR #130, opened 2019-06-27
Construct version from annotated tags only

mrvdb:
Fixes issue #125 

Use `--tags` to use annotated tags only for version construction.

---

- [X ] I have signed the [CLA](https://phabricator.write.as/L1)


thebaer:
Perfect, thanks!

PR #138, opened 2019-07-14
Don't consider post unpublished when title exists

thebaer:
Previously, you could create a post with a title but no body, e.g. by publishing via email, and it would still show the post on a blog but give a 410 Gone page when trying to access the page. This fixes that.

This issue originally reported [on the forum](https://discuss.write.as/t/removing-post-unpublished-by-author-post/725)


thebaer:
Thanks for the review, @robjloranger. Merging now.

PR #142, opened 2019-07-21
Automatic certificates from Let's Encrypt

thebaer:
This enables admins to run WriteFreely as a standalone server that automatically generates certificates, powered by Let's Encrypt ([T542](https://writefreely.org/tasks/542)).

It adds a new config option under the `[server]` section: `autocert`. When set to `true` (and with non-empty `tls_cert_path` and `tls_key_path` values, plus `port = 443`), WriteFreely will serve the instance on port 443 and automatically generate a certificate for the configured host.

This also adds a new option during the interactive config process that allows the user to choose automatic certificate generation, in addition to the previous option of manually configuring the certificate location.

### DIFF
--- a/app.go
+++ b/app.go
@@ -382,16 +382,6 @@ func Serve(app *App, r *mux.Router) {
 	}
 	var err error
 	if app.cfg.IsSecureStandalone() {
-		log.Info("Serving redirects on http://%s:80", bindAddress)
-		go func() {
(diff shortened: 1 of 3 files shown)


thebaer:
This still needs documentation.

thebaer:
Ideally we'll also explain how to do this in some kind of "setup" guide. But the basic docs are there, so merging now.

PR #143, opened 2019-07-21
Style collection 404 page like rest of blog

thebaer:
This displays the "page is missing" text within the same page as any other blog post, keeping customizations, pinned pages, and general blog navigation.

[T493](https://writefreely.org/tasks/493)


thebaer:
Still TODO:

* Return better responses for different requested `Content-Type`s
* Show a similar page when there are no posts with a given tag (probably out of scope for this specific task)

thebaer:
The first item is fixed now, and the second is now tracked in [T682](https://writefreely.org/tasks/682).

PR #144, opened 2019-07-22
Fix missing hostname when publishing via API

thebaer:
This fixes a bug that occurred only when publishing via API and authenticating via token (rather than cookie). I noticed this while testing the new `wf-cli`.

Previously, the instance's hostname wouldn't be added to the `Collection` that got passed around after retrieving the owned post, meaning an incomplete URL was returned in the API response, and federation failed due to the missing host (and thus failed key lookup).


thebaer:
Leaving this open for a bit longer while we test `wf-cli`.

thebaer:
Good to go now.

PR #145, opened 2019-07-22
Fix single-user instance issues

thebaer:
While running a single-user instance on https://matt.writefreely.dev I've noticed some (mostly superficial) issues, and [documented them here](https://matt.writefreely.dev/single-user-issues-so-far):

- [x] “Public” option shows on blog Customize page (should never be there in this mode)
- [x] Pinning a post links to /matt/about instead of /about
- [x] Instance-level About page shows at the bottom of user pages (should never show in this mode)
- [x] Drafts page shows “Find your blog posts from the Blogs page.” (should never show in this mode)

This PR aims to fix those issues.


thebaer:
Fixed pinned posts and haven't heard about any other single-user issues, so merging now.

PR #148, opened 2019-08-01
Support setting default blog visibility on instance

thebaer:
This adds a new `default_visibility` config value that lets an instance admin set the visibility of newly created collections: Unlisted (current default), Public, or Private.

This can also be set via the Admin UI.

Resolves [T675](https://writefreely.org/tasks/675).


thebaer:
Need to verify that creating a collection via API still works, then we can merge.

thebaer:
Verified collection creation via API works, refactored some things, and added documentation. Merging now.

PR #151, opened 2019-08-01
Let Apper set request log format

thebaer:
This adds a new `ReqLog()` func to the `Apper` interface that'll return the log message for incoming requests.

Resolves [T649](https://writefreely.org/tasks/649)


thebaer:
This is pretty straightforward and doesn't change any behavior, so merging now.

PR #153, opened 2019-08-05
Add editor config option

thebaer:
This adds a new `editor` value in the `[app]` config section, which lets admins choose the editor template file to load for all users.

Resolves [T677](https://writefreely.org/tasks/677). See #155 for first alternative editor.


thebaer:
Going to skip documentation for now, until there are finalized editors to choose from. Merging.

PR #154, opened 2019-08-05
Simplified navigation

thebaer:
This provides the option for a more intuitive navigation scheme, prioritizing familiarity for writers over our normal scheme that optimizes for writing. It resolves [T680](https://writefreely.org/tasks/680) and builds on #153 / [T677](https://writefreely.org/tasks/677).

This is a large work in progress and currently experimental. Expect it to change a lot before merging.


thebaer:
This config value might disappear with / be rolled into #157.

thebaer:
This will definitely be renamed / changed around a bit. Closing and rolling this into #157.

PR #157, opened 2019-08-06
Reader-first multi-user instances

thebaer:
This allows instance admins to configure an instance to be more about the collective Reader than the individual blogs.

To enable, edit your configuration file to have `chorus = true` under the `[app]` section. This is still very experimental, is built to solve a specific user's needs, and will likely change in drastic ways in the near future. It builds off the work in #153, #154, and #155, and will resolve [T681](https://writefreely.org/tasks/681) when finished.


thebaer:
@robjloranger this is ready for review now. Again, this is mostly for our own purposes (things like our [internal WF instance](https://team.write.as)), so there are still some missing / incomplete features here.

All new config options are considered **experimental**, so we might not address all issues with different combinations of config values -- that'll be for a later iteration. For now, I'm mostly interested in verifying that this doesn't introduce any regressions, and everything runs / builds correctly for you.

Final list of changes in this PR:

* New `editor` choice: `bare`. This editor provides a very pared-down writing experience, and is built for multi-user, 1-blog instances.

* New `simple_nav` option. This adds a prominent "Log out" and "New post" button. It's meant to be used with the `chorus` option.

* New `chorus` option. This changes an instance to be focused on _reading_ first, rather than _writing_ first. It includes more consistent site-wide navigation, whether logged in or not, and new collection templates that downplay the individual author. Also when enabled, hashtags link to instance-wide tagged posts on the Reader instead of user-only tagged posts on their individual blog.

* Customizable Reader page. Admins can now edit the Reader page from the dashboard. The Reader page design has been changed to be more consistent with other pages (About, Privacy) to help facilitate this.

* Fix login bug. Previously, a failed login attempt would change around the site-wide navigation on the `/login` page to look as if a user was logged in, even though they weren't. That's now fixed.

thebaer:
Thanks for taking a look at everything, and glad you like it!

> The new `Post` button is only present when this is set, so it would be more consistent.

Which `Post` button do you mean?

> I was wondering if it looks better to remove the Logout and separator from the user drop down menu when this is set to true, as there is now a logout link in the header.

This is a great point. But I think for now, while we're still trying to get the interface right for the non-technical users choosing this layout, the redundancy is okay.

robjloranger:
Oh the new simple NAV style post button on the pad.

thebaer:
Ah, yeah it's styled the same but it's actually unrelated to either of those settings -- it's just a static part of the `bare` editor. (Though I suppose in the future we could do that -- put that style of button in all editor templates and have it respect the `simple_nav` setting.)

PR #163, opened 2019-08-09
Customize Reader page

thebaer:
This makes it possible to edit the title and introductory text at the top of the Reader view.

Resolves [T684](https://writefreely.org/tasks/684)


thebaer:
Let me do this another way.

PR #164, opened 2019-08-09
fix tar bombs

### LINKED ISSUE #160: Tar is bombing the cwd with files
### Describe the bug
Untarring the writefreely release with the customary `tar xzvf writefreely_x.x.x_os_arch.tar.gz` explodes littering the current directory with files. The tar should have a root directory so that untarring creates only one directory in the cwd.

1. Download writefreely
2. `tar xzvf writefreely_x.x.x_os_arch.tar.gz` 
3. The contents litter the current directory

### Expected behavior
A single directory named `writefreely` should be extracted with the rest of the files placed inside it

**Version or last commit**: 0.10.0

robjloranger:
this changes the release targets in the Makefile to use a subdirectory
of the format BINARYNAME_GITREV so extracting the archive results in a
single directory.

---

closes #160

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,7 @@ GOBUILD=$(GOCMD) build $(LDFLAGS)
 GOTEST=$(GOCMD) test $(LDFLAGS)
 GOGET=$(GOCMD) get
 BINARY_NAME=writefreely
+TAR_NAME=$(BINARY_NAME)_$(GITREV)
 DOCKERCMD=docker
 IMAGE_NAME=writeas/writefreely
 TMPBIN=./tmp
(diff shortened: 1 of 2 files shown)


thebaer:
What do you think if we name the directory `writefreely` instead of including the version info? That should make installation / upgrading easier by maintaining existing paths, e.g.:
```
tar -C /var/www -xzf writefreely_0.10.0_linux_amd64.tar.gz
```
always produces `/var/www/writefreely`.

joyeusenoelle:
The only problem I can think of is a rare edge case where someone wants multiple versions installed, but they can do that manually so I'm not sure we need to care about it.

robjloranger:
sure, sounds good to me

thebaer:
That fixes it! Looks good -- thanks @robjloranger.

PR #165, opened 2019-08-09
check for lessc executable in any location

robjloranger:
previously the checks were explicit locations which does not work when
using something like nvm to manage node packages and versions.

this checks for the executable and sets the script variable LESSC to the
full path of the one found.
if none was found the make command will error.

---

### DIFF
--- a/less/Makefile
+++ b/less/Makefile
@@ -1,9 +1,11 @@
 CSSDIR=../static/css/
 
 all :
-	ifeq (,$(shell command -v lessc 2> /dev/null))
-	$(error "less is not installed, please run install-less.sh")
-	endif
+ifeq (,$(shell command -v lessc 2> /dev/null))
+$(error "less is not installed, please run install-less.sh")
+endif
+	lessc app.less --clean-css="--s1 --advanced" $(CSSDIR)write.css
+	lessc fonts.less --clean-css="--s1 --advanced" $(CSSDIR)fonts.css
 	lessc icons.less --clean-css="--s1 --advanced" $(CSSDIR)icons.css
 
 install :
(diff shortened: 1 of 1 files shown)


thebaer:
When running `make install` from the `less` dir I got:

```
Makefile:4: *** recipe commences before first target.  Stop.
```

So I changed the file a bit, and that seemed to fix it. Let me know if it messes anything up when you try it.

robjloranger:
works for me still as well :+1:

thebaer:
Great, and just to confirm, it works when `lessc` isn't previously installed?

robjloranger:
hmm, no it doesn't seem to detect the binary is missing.

robjloranger:
this one is weird, I guess the variable is never undefined but not always what we want?

thebaer:
Hmm, yeah... Maybe an empty check instead would do it?

robjloranger:
white space was throwing me around, double check this one for me

thebaer:
It seems to be catching it, but now `make install` doesn't work..

robjloranger:
weird, what is the error? it works .. on my machine :rofl:

thebaer:
Heh :smile: Error is:

```
Makefile:2: *** "less is not installed, please run install-less.sh".  Stop.
```

Instead, `make install` should run `install-less.sh`, etc.: https://github.com/writeas/writefreely/blob/66974dcbffb498456523f0fed243052289bfdf22/less/Makefile#L19-L21

robjloranger:
oooh, `make install` in `less` I get it. fixed? now :grin:

thebaer:
Great, now let's just make sure all the stylesheets are still generated :)

robjloranger:
this is a mess, I fix one issue and the other reappears. Maybe we need to use some bash instead.

robjloranger:
ok, ready for another test @thebaer

PR #166, opened 2019-08-09
fix panic on duplicate remoteuser key

### LINKED ISSUE #159: When Alice (a fediverse actor) tries to follow a writefreely actor (Bob) and Alice already follows Bob then writefreely crashes.
### Describe the bug
When a follow is already registered in the writefreely database, repeating the follow request crashes writefreely

1. Follow a writefreely author from an instance you control
2. Fool your instance to think step 1 never happened (remove the follow from the database)
3. Repeat 1
4. Observe writefreely crash

### Expected behavior
Nothing probably. Or maybe refresh the public key with the new one?

### Application configuration
 - **Single mode** 
 - **sqlite** [mysql/sqlite]
 - **Open registration?** no
 - **Federation enabled?** yes

**Version or last commit**: 0.10.0

[log of installation, execution and crash with --debug](https://github.com/writeas/writefreely/files/3480381/writefreely.log)

robjloranger:
this changes handleFetchCollectionInbox to log _all_ errors after
attempting to insert an actor in the remoteusers table. previously
checking for all errors _except_ duplicate keys would cause a panic if
an actor made a request to follow while already having followed.

---

closes #159

### DIFF
--- a/activitypub.go
+++ b/activitypub.go
@@ -375,11 +375,11 @@ func handleFetchCollectionInbox(app *App, w http.ResponseWriter, r *http.Request
 				// Add follower locally, since it wasn't found before
 				res, err := t.Exec("INSERT INTO remoteusers (actor_id, inbox, shared_inbox) VALUES (?, ?, ?)", fullActor.ID, fullActor.Inbox, fullActor.Endpoints.SharedInbox)
 				if err != nil {
-					if !app.db.isDuplicateKeyErr(err) {
-						t.Rollback()
-						log.Error("Couldn't add new remoteuser in DB: %v\n", err)
-						return
-					}
+					// if duplicate key, res will be nil and panic on
+					// res.LastInsertId below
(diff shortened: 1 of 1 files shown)


robjloranger:
to test the fix use a local isntance and try something like the following:
```
curl -H "Content-Type: application/json" \
    --request POST \
    --data '{
    "@context": "https://www.w3.org/ns/activitystreams",
    "id": "https://my-example.com/my-first-follow",
    "type": "Follow",
    "actor": "https://mastodon.technology/@robjloranger",
    "object": "https://mastodon.social/users/Mastodon"
    }' \
    http://localhost:8080/api/collections/rob/inbox
```

The collection in the target inbox needs to exist and the actor may be any activity pub actor that is live.

thebaer:
Wasn't able to replicate the crash with this (or a few other things I tried), but this should at least fix it. Thanks! Merging now.

PR #168, opened 2019-08-10
Emit the server software and version to the log on startup

OddBloke:
As suggested (by me) in https://discuss.write.as/t/log-the-writefreely-version-on-startup/763

This follows the same format as `writefreely -v`, so also refactors `OutputVersion()` to use a new `FormatVersion()` function.

---

### DIFF
--- a/app.go
+++ b/app.go
@@ -486,9 +486,14 @@ func ConnectToDatabase(app *App) error {
 	return nil
 }
 
+// FormatVersion constructs the version string for the application
+func FormatVersion() string {
+	return serverSoftware + " " + softwareVer
+}
+
 // OutputVersion prints out the version of the application.
 func OutputVersion() {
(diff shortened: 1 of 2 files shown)


thebaer:
Thanks for contributing this! Good idea, and simple enough fix.

My only feedback is that I think the `log.Info()` call should happen inside `main()` (in `cmd/writefreely/main.go`) instead of `writefreely.Initialize()`. That would keep the `writefreely` package flexible, and prevents logging the wrong software name / version if it's ever used for other kinds of executables. What do you think?

OddBloke:
@thebaer Thanks for the review! I agree, and have made the change.

OddBloke:
Anything else I can do to push this forward?

thebaer:
Just wanted to run it myself -- all good! Thanks for contributing this!

OddBloke:
Thanks!

PR #169, opened 2019-08-12
prevent future posts from showing in pins

### LINKED ISSUE #158: Pinned scheduled posts are visible to all users before publish date
### Describe the bug
Pinning a future-dated post will make it show for all users, when it should only show for the author of the blog.

1. Publish a post
1. Change publish date to the future
1. Pin the post
1. View the blog in another browser
1. Notice the post is pinned to the navigation, and clicking it shows the full post (with "Scheduled" badge)

### Expected behavior
The post shouldn't show up in blog navigation menu until the published date has passed.

### Application configuration
 - **Single mode or Multi-user mode?** either
 - **Database?** either
 - **Open registration?** either
 - **Federation enabled?** N/A

**Version or last commit**: v0.10.0

robjloranger:
this changes GetPinnedPosts to accept an includeFutre bool, which
returns future dated pinned posts when true.

---

closes #158

### DIFF
--- a/collections.go
+++ b/collections.go
@@ -783,11 +783,7 @@ func handleViewCollection(app *App, w http.ResponseWriter, r *http.Request) erro
 
 	// Add more data
 	// TODO: fix this mess of collections inside collections
-	if isOwner {
-		displayPage.PinnedPosts, _ = app.db.GetPinnedPosts(coll.CollectionObj, true)
-	} else {
-		displayPage.PinnedPosts, _ = app.db.GetPinnedPosts(coll.CollectionObj, false)
-	}
+	displayPage.PinnedPosts, _ = app.db.GetPinnedPosts(coll.CollectionObj, isOwner)
 
 	err = templates["collection"].ExecuteTemplate(w, "collection", displayPage)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


collections.go
```diff
@@ -782,7 +783,11 @@ func handleViewCollection(app *App, w http.ResponseWriter, r *http.Request) erro
 
 	// Add more data
 	// TODO: fix this mess of collections inside collections
-	displayPage.PinnedPosts, _ = app.db.GetPinnedPosts(coll.CollectionObj)
+	if isOwner {
+		displayPage.PinnedPosts, _ = app.db.GetPinnedPosts(coll.CollectionObj, true)
+	} else {
+		displayPage.PinnedPosts, _ = app.db.GetPinnedPosts(coll.CollectionObj, false)
```
thebaer: We can simplify this kind of logic back to one line, and just use `isOwner` for the second parameter.

database.go
```diff
@@ -1533,9 +1533,15 @@ func (db *datastore) GetLastPinnedPostPos(collID int64) int64 {
...
-func (db *datastore) GetPinnedPosts(coll *CollectionObj) (*[]PublicPost, error) {
+func (db *datastore) GetPinnedPosts(coll *CollectionObj, includeFuture bool) (*[]PublicPost, error) {
 	// FIXME: sqlite-backed instances don't include ellipsis on truncated titles
-	rows, err := db.Query("SELECT id, slug, title, "+db.clip("content", 80)+", pinned_position FROM posts WHERE collection_id = ? AND pinned_position IS NOT NULL ORDER BY pinned_position ASC", coll.ID)
+	rows := &sql.Rows{}
+	var err error
+	if includeFuture {
+		rows, err = db.Query("SELECT id, slug, title, "+db.clip("content", 80)+", pinned_position FROM posts WHERE collection_id = ? AND pinned_position IS NOT NULL ORDER BY pinned_position ASC", coll.ID)
+	} else {
+		rows, err = db.Query("SELECT id, slug, title, "+db.clip("content", 80)+", pinned_position FROM posts WHERE collection_id = ? AND pinned_position IS NOT NULL AND created <= "+db.now()+" ORDER BY pinned_position ASC", coll.ID)
```
thebaer: I think we should keep this in one `db.Query()` call and append the time condition if needed, exactly like we do in `GetPinnedPosts()`. That'll keep things consistent and help prevent potential bugs from any future changes to the query.

robjloranger:
concerns addressed

thebaer:
Looks great, thanks! I just made a small change, now using the new `isOwner` var in the `if` statements. Merging now.

PR #171, opened 2019-08-12
fix missing collection hostname

### LINKED ISSUE #161: Collection.hostName is empty in the logs
No UI behaviour mishap as far as I can see.

Upgrading to revision `8557119451f49638e69dc128a0842bd09e7dc9c9` on `develop` this line starts showing up in the logs and instructs to file a bug.

robjloranger:
GetCollections and GetPublishableCollections now take a hostname
parameter to allow setting the collecion hostname.

All collections used in memory now have their hostname set.

---

closes #161

### DIFF
--- a/account.go
+++ b/account.go
@@ -13,6 +13,13 @@ package writefreely
 import (
 	"encoding/json"
 	"fmt"
+	"html/template"
+	"net/http"
+	"regexp"
+	"strings"
+	"sync"
+	"time"
+
 	"github.com/gorilla/mux"
 	"github.com/gorilla/sessions"
 	"github.com/guregu/null/zero"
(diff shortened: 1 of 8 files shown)


robjloranger:
This does seem like a lot of extra work compared to storing the hostname in the collection. The post could then get it's hostname from it's associated collection and we wouldn't be passing it around.

Then for things like export, we could omit the hostname value to avoid issues moving blogs across deployments.

But that is for another day, this fixes the mentioned issue.

thebaer:
Yeah, I'm just not excited about adding a variable to the `Collection` struct that isn't going to change between different `Collection`s. This seems like the least-worst way to do things, at least.

But looks good, thanks!

PR #172, opened 2019-08-16
add basic text file imports

robjloranger:
this adds basic support for importing files as blog posts.

.txt and .md are supported at this time and the
collection is selectable, defaulting to draft.

if a collection is specified the post is federated.

[T609](https://phabricator.write.as/T609)

---


robjloranger:
I got a big sidetracked down the rabbit hole of #96, once I figured out what was getting in my way I remembered seeing something about the `NOW` db function and found the PR.

so there may be things I missed migrating to a fresh branch, but quick manual tests were successful.

robjloranger:
It's looking pretty good. I would prefer if the errors could be in a similar style box as the success alert, it would look a bit more polished.

Other than that I might change the form to allow all plain text files, not just .txt and .md. Then verify they are plain text again on the server. Not everyone will use files extensions.

account_import.go
```diff
@@ -0,0 +1,155 @@
...
+			Font:    "norm",
+		}
+		if coll.ID != 0 && app.cfg.App.Federation {
+			token, err := app.db.GetAccessToken(u.ID)
+			if err != nil {
+				fileErrs = append(fileErrs, fmt.Errorf("failed to authenticate uploading: %s", formFile.Filename))
+				log.Error("import textfile: get accesstoken: %+v", err)
+				continue
+			}
+			ownedPost, err := app.db.CreateOwnedPost(&submittedPost, token, coll.Alias, app.cfg.App.Host)
```
thebaer: I think we should be calling `CreatePost()` here, too. I know the whole setup is a bit messy, but `CreateOwnedPost()` was meant for direct API usage, and if we use it here, we're going to fill up the database with access tokens that are only used once (and waste resources generating those tokens).

If this route was taken in order to get a `PublicPost` back, we should just duplicate the `PublicPost` creation logic from `CreateOwnedPost()` instead of going through token generation.

routes.go
```diff
@@ -93,6 +94,8 @@ func InitRoutes(apper Apper, r *mux.Router) *mux.Router {
 	me.HandleFunc("/posts/export.json", handler.Download(viewExportPosts, UserLevelUser)).Methods("GET")
 	me.HandleFunc("/export", handler.User(viewExportOptions)).Methods("GET")
 	me.HandleFunc("/export.json", handler.Download(viewExportFull, UserLevelUser)).Methods("GET")
+	me.HandleFunc("/import", handler.User(viewImport)).Methods("GET")
+	me.HandleFunc("/import", handler.User(handleImport)).Methods("POST")
```
thebaer: In general, we should try to keep endpoints that perform actions on the `/api/` subroute. So I think this should be on `/api/me/import` instead of `/me/import`

thebaer:
Also, good idea on allowing all plain text files, and not just those with a file extension.

robjloranger:
Awesome, I'll fix these up shortly.

I think I tried using `CreatePost` at first but ran into trouble with something, which may have been my doing.

robjloranger:
@thebaer all set, tested on my end and still working

thebaer:
Thanks for getting those things! After testing, I made a few changes.

I tweaked the design and copy on the Import page to be consistent with the rest of the site.

I also ran into problems when importing Draft posts -- whether from bad local data or something else, the `GetCollection("")` call actually returned a collection my account didn't own, and the posts were imported to it anyway. So I fixed the logic there to protect against this and to be more consistent with how we do these draft / collection post checks elsewhere in the app. And I updated the logic to make sure you can't import posts to collections you don't own.

I also reduced the number of database calls (previously, this was repeatedly making the same `GetCollection()` call in the `for` loop). And I fixed issues with putting `defer` calls for file operations in a loop.

Next I'll probably tweak the stuff around temporary files a bit.

account_import.go
```diff
@@ -0,0 +1,186 @@
...
+		} else if err != nil {
+			fileErrs = append(fileErrs, fmt.Errorf("failed to read copy of %s", formFile.Filename))
+			log.Error("import textfile: file to post: %v", err)
+			continue
+		}
+
+		if collAlias != "" {
+			post.Collection = collAlias
+		}
+		created := post.Created.Format("2006-01-02T15:04:05Z")
```
thebaer: Generating a temporary file from the upload will mean the `Created` time is always _now_. Instead, we should use the client-side [File API](https://developer.mozilla.org/en/docs/Web/API/File) to send the original file creation time along with each file, and then set `post.Created` to that form value.

robjloranger:
I had to use the last modified time as the File API does not provide created

thebaer:
Cool, that's not a problem (and maybe better, anyway). Testing now.

heyakyra:
What information is supported with imports (authors, tags, etc.)? Is there official documentation on this somewhere?

thebaer:
@heyakyra, built-in text file imports only support:

* title
* body
* publish date

For more comprehensive imports, the API is probably the best way to migrate ([documentation here](https://developers.write.as/docs/api/))

PR #173, opened 2019-08-21
Import zips

ghost:
This is just a draft for early discussion, it is dependent on #172 and will likely need changes.

Expect force pushes while work continues on text imports.

---


robjloranger:
the only thing really missing, other than pretty front end things, is importing just .txt files. that's a quick change though, a new `ZipFunc` to filter on only `.txt` and use `FromZipDirsFunc` instead.

joyeusenoelle:
Hang on, Rob, we may be duplicating effort. I believe I have zip imports working in https://github.com/writeas/wp-import/tree/writefreely-import .

robjloranger:
We have indeed, rather I have. I should have jumped over there before writing the zip part.

I think we can merge the general purpose zip parsing logic into [writeas/import](https://gitub.com/writeas/import), not right away but eventually. My goal with `import` was to have very general purpose utilities to get posts from different files.

Then we will have a library for parsing posts from various files and formats. We can still then have a `wp-import` tool and also use it in writefreely itself. Then later we could add import functionality into the CLI as well, and others can create their own tools if they wish.

joyeusenoelle:
That was my eventual goal as well, if I'm understanding you correctly, so we're on the same page.

A note, while we're talking about it: although Medium's export format is ZIP, it needs a little more finagling than Write.as/WriteFreely's ZIPs, since there's a bunch of stuff in there we don't care about and the posts are in HTML format.

Do we have existing code, or a recommended library, for converting HTML to markdown? I don't remember offhand.

robjloranger:
:+1: 

In the import library there is a function `FromZipDirsByFunc` which allows filtering based on any logic you wish, including parsing HTML to markdown, or skipping certain files.

So we could add, say `FromZipMedium` or similar, and that could use a `ZipFunc` to filter and process accordingly.

AFAIK there is not any existing code we have for conversion, there is probably a nice library somewhere.

thebaer:
Going to close this, since this is old and much has changed. We'll still support this through our new import system.

PR #179, opened 2019-09-07
set a default pad tempate on all pad renders

robjloranger:
this fixes a bug where if the `editor` config is set to an unsupported
value there is a nil pointer error and the pad fails to render when
editing only, not on a new post.

---


thebaer:
Great catch -- thanks!

Issue #176, opened 2019-09-01
v0.10.0 release binaries for linux ships without MathJax 


RJ722:
### Describe the bug

The [linux release binary v0.10.0][bin] ships with an empty MathJax directory. Therefore, no math is rendered on a blog despite turning on MathJax.

This issues was originally reported on [discourse](https://discuss.write.as/t/new-formatting-syntaxes/43/15?u=rj722).

[bin]: https://github.com/writeas/writefreely/releases/download/v0.10.0/writefreely_0.10.0_linux_amd64.tar.gz

### Steps to reproduce (if necessary)
1. Untar [linux release binary v0.10.0][bin]
2. the `static/js/mathjax` directory is empty.

### Application configuration
 - **Single mode or Multi-user mode?** Multi-user mode
 - **Database?** mysql
 - **Open registration?** no
 - **Federation enabled?** yes

**Version or last commit**:

v0.10.0

robjloranger:
Thanks for this @RJ722, looks like our update script needs to account for using a submodule for MathJax. Sorry for the inconvenience.

robjloranger:
Hey sorry for the delay here, I will work on this change today.

thebaer:
I've updated the [v0.10 builds](https://github.com/writeas/writefreely/releases/tag/v0.10.0) to all include Mathjax, so this particular issue is resolved now.

We'll continue work on optimizing this on #181.

PR #181, opened 2019-09-11
Mathjax in binaries

robjloranger:
removes the submodule for mathjax and instead includes a copy of the script

ref #176, will need a tag and new release to close

---


thebaer:
I think we also need the fonts and images from the library, right?

robjloranger:
I'm not certain, the docs just say to link the one script.

I did notice that the body is not being parsed properly though so I will need to see what's going on here.

robjloranger:
this uses the svg version which does not require any external fonts
Screenshot from 2019-09-11 14-17-28

thebaer:
I'm not sure how to verify this since I'm not really a user of this feature. If anyone in the community can try it out and weigh in on whether or not it works, please do!

robjloranger:
Hey @thebaer, to verify this is working you can try somewhere like http://atomurl.net/math/ and use a sample from the selection. This will give you the markup and a preview of it rendered to compare to.

thebaer:
Worked for me! Merging now.

PR #183, opened 2019-09-13
add user invite instructions

robjloranger:
this adds a new page with instructions for sharing user invites

if a user clicks the link for one of their own invite codes they are
directed to a page with clear instructions for it's use.

if a user clicks another users link they are redirectec to their account
settings witha flash telling them they do not need to register.

ref [T690](https://writefreely.org/tasks/690)

---

### DIFF
--- a/database.go
+++ b/database.go
@@ -2257,6 +2257,19 @@ func (db *datastore) GetUserInvite(id string) (*Invite, error) {
 	return &i, nil
 }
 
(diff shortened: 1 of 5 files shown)


robjloranger:
let me know if `/me/settings` will not work as a redirect for logged in users trying to use an invite code from another user. I ran the `chorus` branch with it enabled and the endpoint didn't appear to be changed.

PR #184, opened 2019-09-18
allow titles for abbreviation elements

robjloranger:
this allows abbreviation elements to keep their title attributes when
containing special characters.

---


robjloranger:
I double checked this allow rule still prevents common XSS attack vectors. Like closing the title attribute and inserting an onload, or closing the element entirely and injecting a new element like `script`. This either omits the abbreviation element entirely or just renders some of the contents as text on the page. Harmless either way.

But now titles such as "This is a title: with a sub tile" render as expected.

PR #189, opened 2019-10-02
T697 default to user's first blog instead of draft

robjloranger:
This changes the pad `publish to` menu to default to the user's first blog, unless a previous publish target was used and moves the `Drafts` selector to the bottom of the list.

---

ref [T697](https://writefreely.org/tasks/697)


thebaer:
Code looks good after reading through -- just need to run and verify.

PR #190, opened 2019-10-02
add dates to posts on collections w/ blog format

robjloranger:
This adds conditional date placement on collection posts using the 'blog' format.

When dates are shown without a post title, the date is formatted as the body text with top placement.

---

ref T669

### DIFF
--- a/less/core.less
+++ b/less/core.less
@@ -86,6 +86,7 @@ body {
 		}
 
 		header {
+			padding: 1em 1rem;
 			nav {
 				span, a {
 					&.pinned {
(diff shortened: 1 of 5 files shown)


robjloranger:
I changed the chorus template as well, quick to revert if you prefer. I thought it would be good to be consistent

thebaer:
Thanks for giving this a shot, but a lot of the layout got messed up here, and it looks like it changes the previous header spacing we had (which ideally we'd avoid). So I took another stab at this in #235. If that works, we'll merge that instead.

PR #191, opened 2019-10-03
show timestamps in local date/locale

robjloranger:
this adds a helper script to rewrite all time elements with a proper
datetime attribute into the users locale via the browser
navigator.language.

collection, collection-post and chorus-collection-post templates now
include this script

---

requires #190, might need a rebase

ref [T670](https://writefreely.org/tasks/670)


thebaer:
Let's also get this code on the `templates/chorus-collection.tmpl` template, collection tags templates, and the Reader view.

thebaer:
Thanks :+1: Is this PR ready to go or still a draft?

robjloranger:
It's good, #190 will just need the script included after.

thebaer:
Also needed to handle cases where posts don't have a title (so the date becomes the clickable title on blog index, tag page). Previously, the script would overwrite the link so you couldn't navigate to the post itself on those pages. All fixed and good to go! Thanks!

PR #192, opened 2019-10-03
T695 reset user pass

robjloranger:
this adds a new password reset feature the the admin user view page, passwords are randomly generated

---

ref [T695](https://phabricator.write.as/T695)


thebaer:
Tweaked some copy and design, and made it so the admin can see user's email address (if they have one) in order to get in touch. This also uses [`passgen.NewWordish()`](https://godoc.org/github.com/writeas/web-core/passgen#NewWordish), instead, for password generation.

PR #194, opened 2019-10-08
lightly style tables in posts

robjloranger:
styling added:
- thin inner borders
- table body rows alternate light background
- table header text centered
- cellspacing reset and replaced with padding to avoid gaps in borders

---

### DIFF
--- a/less/core.less
+++ b/less/core.less
@@ -632,6 +632,30 @@ table.classy {
 	}
 }
 
+body#post article table {
+	border-spacing: 0;
+	border-collapse: collapse;
+	th {
+		border-bottom: 2px solid #ccc;
+		text-align: center;
+	}
+	th + th {
+		border-left: 1px solid #ccc;
+	}
+	tr:nth-child(even) {
+		background-color: #f6f6f6;
+	}
+	td + td {
+		border-left: 1px solid #ccc;
+	}
+	tr + tr td {
+		border-top: 1px solid #ccc;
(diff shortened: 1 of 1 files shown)


thebaer:
This looks nice, but I think our default style should be as simple as possible, so it's as easy as possible to override with custom CSS. Then we can make this available as a fancier option.

Here, I think we should just make the table `width: 100%` and add some light borders.

robjloranger:
How does this look? Or were you thinking outter borders as well? (ignore the top/left lines, screencap noise)

2020-02-11-085744_707x247_scrot

thebaer:
Looks good, but is the CSS simple / easily to override?

robjloranger:
2020-02-11-094647_675x216_scrot

I reworked it a bit to make it simpler to override. To change the style would require something like:

```css
article table {
  border-color: new-color;
}

article table th {
  font-size: 3em;
}
```

We could change to using a class but it would be similar to override like:

```css
table.post {
  border-color: new-color;
}

table.post th {
  font-size: 3em;
}
```

thebaer:
Cool, I think that looks great :+1:

cjeller1592:
Tables look good on my end! Both Chrome (80.0.3987) and Firefox (73.0.1) show the same table styling.

thebaer:
Thanks @robjloranger and @cjeller1592! Looks good, merging now.

PR #195, opened 2019-10-09
Send out ActivityPub mentions (T627)

qwazix:
This pull request adds code that scans the text of each post to be published for handles of the type @username@domain.tld, uses webfinger to ask ActivityPub endpoints about them, gets their inbox and actor IRI's and adds mention tags to the post. Finally it adds the users to the CC list of the recipients of the activity.

This works in _pleroma_ but not in _mastodon_ unless we set the object type to "Note" instead of "Article"

[T627](https://phabricator.write.as/T627)


thebaer:
Latest changes look good :+1: Confirmed that mentions work with Pleroma.

sqlite.sql
```diff
@@ -172,6 +172,7 @@ CREATE TABLE IF NOT EXISTS `remoteusers` (
   actor_id TEXT NOT NULL,
   inbox TEXT NOT NULL,
   shared_inbox TEXT NOT NULL,
+  handle TEXT DEFAULT '' NOT NULL,
```
thebaer: No need to add this to the schema files, as the migrations run on install and take care of this.

schema.sql
```diff
@@ -181,6 +181,7 @@ CREATE TABLE IF NOT EXISTS `remoteusers` (
   `actor_id` varchar(255) NOT NULL,
   `inbox` varchar(255) NOT NULL,
   `shared_inbox` varchar(255) NOT NULL,
+  `handle` varchar(255) DEFAULT '' NOT NULL,
```
thebaer: No need to add this to the schema files, as the migrations run on install and take care of this.

collections.go
```diff
@@ -820,6 +820,22 @@ func handleViewCollection(app *App, w http.ResponseWriter, r *http.Request) erro
...
+	vars := mux.Vars(r)
+	handle := vars["handle"]
+
+	remoteUser, err := app.db.getProfilePageFromHandle(app, handle)
+	if err != nil {
+		log.Error("Couldn't find this user "+handle, err)
+		return nil
+	}
+
+	http.Redirect(w, r, remoteUser, http.StatusSeeOther)
```
thebaer: We should use our common pattern of `return`ing an `impart.HTTPError` instead of calling this method (for example): https://github.com/writeas/writefreely/blob/bc2016f00ff2b2fa58f978f489255457af4f46d5/collections.go#L416

The `handler.Web()` method that wraps this takes care of the actual redirect.

activitypub.go
```diff
@@ -588,36 +589,73 @@ func federatePost(app *App, p *PublicPost, collID int64, isUpdate bool) error {
...
+			activity.To = na.To
+			activity.CC = na.CC
+			// This here might be redundant in some cases as we might have already
+			// sent this to the sharedInbox of this instance above, but we need too
+			// much logic to catch this at the expense of the odd extra request.
+			// I don't believe we'd ever have too many mentions in a single post that this
+			// could become a burden.
+
+			fmt.Println(tag.HRef)
+			fmt.Println("aaa")
```
thebaer: Just need to clean these up (and the comment below)

qwazix:
Ack to all, will do.

thebaer:
Also, just saw the comments on the Mastodon issue. Is that fixed on our end?

qwazix:
Yes it works as expected, Gargron was right, I was thrown off because the first alias that webfinger returned was the @ version.

routes.go
```diff
@@ -184,6 +184,7 @@ func InitRoutes(apper Apper, r *mux.Router) *mux.Router {
 
 func RouteCollections(handler *Handler, r *mux.Router) {
 	r.HandleFunc("/page/{page:[0-9]+}", handler.Web(handleViewCollection, UserLevelReader))
+	r.HandleFunc("/mention:{handle}", handler.Web(handleViewMention, UserLevelReader))
```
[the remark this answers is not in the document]
qwazix: Done. I put that in `initRoutes`, is that the right place?

sqlite.sql
qwazix: Reverted to previous

schema.sql
qwazix: Reverted to previous

collections.go
qwazix: Done

qwazix:
The crash was due to prismo and pixelfed not having sharedInboxes. I made GetSharedInbox() to return the inbox of the user so implementation won't need to care whether there is a sharedInbox or not (this is in activityserve)

routes.go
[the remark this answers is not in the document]
thebaer: Yep, that's the right place :+1:

webfinger.go
```diff
@@ -80,3 +84,46 @@ func (wfr wfResolver) DummyUser(username string, hostname string, r []webfinger.
...
+	}
+
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		log.Error("Error reading webfinger response", err)
+		return ""
+	}
+
+	var result map[string]interface{}
+	json.Unmarshal(body, &result)
```
thebaer: Using `interface{}` with all the forced typecasting below provides a lot of opportunity for the application to crash. Instead, we should `Unmarshal()` into a [`webfinger.Resource`](https://godoc.org/github.com/writeas/go-webfinger#Resource) struct. (See our use of this library in webfinger.go.)

qwazix: Done. The code looks way nicer now too.

If you are unable to visit the post I mentioned you in it's because it's the development writefreely instance running on my laptop and it isn't always online.

thebaer:
Got everything up to date with `develop`; this is almost ready to merge! Last pieces I'm thinking about:

* Maybe using a different path for remote users. Would just like something short and clean.
* Maybe changing the style of mention links. Initial thoughts are to make the domain part of a user's handle faded out a little (but not completely hidden like on Mastodon). I'm also open to any ideas anyone has here

thebaer:
Ah, one more crash caused by [this post](https://matt.writefreely.dev/self-mention):

```
panic: interface conversion: interface {} is nil, not string
goroutine 1153 [running]:
github.com/writeas/activityserve.NewRemoteActor(0xc000d83e00, 0x33, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
        /home/matt/go/pkg/mod/github.com/writeas/activityserve@v0.0.0-20191011072627-3a81f7784d5b/remoteActor.go:37 +0x444
github.com/writeas/writefreely.(*datastore).getProfilePageFromHandle(0xc0006acb40, 0xc000085f40, 0xc00139a44f, 0x1c, 0xc000aaec80, 0x2, 0xa, 0xc000d82bc0)
        /home/matt/go/src/github.com/writeas/writefreely/database.go:2506 +0x11a
github.com/writeas/writefreely.(*PublicPost).ActivityObject(0xc000512240, 0xc000085f40, 0x1)
        /home/matt/go/src/github.com/writeas/writefreely/posts.go:1177 +0xc6c
github.com/writeas/writefreely.federatePost(0xc000085f40, 0xc000512240, 0x1, 0x29fbd00, 0x0, 0x0)
        /home/matt/go/src/github.com/writeas/writefreely/activitypub.go:614 +0xf6
created by github.com/writeas/writefreely.newPost
        /home/matt/go/src/github.com/writeas/writefreely/posts.go:612 +0x81f
```

And the mention links in that post cause a 500 and panic:

```
ERROR: 2019/11/26 19:58:17 handle.go:469: User: devlog
interface conversion: interface {} is nil, not string: goroutine 204 [running]:
runtime/debug.Stack(0xc000085f40, 0xc000579100, 0xc0004dbd80)
        /usr/local/go/src/runtime/debug/stack.go:24 +0x9d
github.com/writeas/writefreely.(*Handler).Web.func1.1.1(0xc0006a6d20, 0xc000579100, 0x1936e40, 0xc0000e01c0, 0xc000515960, 0xbf6f7ed657a0817f, 0x18d536fd42, 0x29fbda0)
        /home/matt/go/src/github.com/writeas/writefreely/handle.go:469 +0x16f
panic(0x161a700, 0xc000628ba0)
        /usr/local/go/src/runtime/panic.go:522 +0x1b5
github.com/writeas/activityserve.NewRemoteActor(0xc0005a2db0, 0x25, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
        /home/matt/go/pkg/mod/github.com/writeas/activityserve@v0.0.0-20191011072627-3a81f7784d5b/remoteActor.go:37 +0x444
github.com/writeas/writefreely.(*datastore).getProfilePageFromHandle(0xc0006a6b40, 0xc000085f40, 0xc00048006d, 0xe, 0xc000443048, 0x403faa, 0x15d7020, 0x191bbb0)
        /home/matt/go/src/github.com/writeas/writefreely/database.go:2506 +0x11a
github.com/writeas/writefreely.handleViewMention(0xc000085f40, 0x1936e40, 0xc0000e01c0, 0xc000579100, 0xc000085180, 0x0)
        /home/matt/go/src/github.com/writeas/writefreely/collections.go:847 +0x9f
github.com/writeas/writefreely.(*Handler).Web.func1.1(0xc0006a6d20, 0xc000579100, 0x1936e40, 0xc0000e01c0, 0x17f67d8, 0x17f6930, 0x0, 0x0)
        /home/matt/go/src/github.com/writeas/writefreely/handle.go:501 +0x166
github.com/writeas/writefreely.(*Handler).Web.func1(0x1936e40, 0xc0000e01c0, 0xc000579100)
        /home/matt/go/src/github.com/writeas/writefreely/handle.go:515 +0x63
net/http.HandlerFunc.ServeHTTP(0xc0006a7180, 0x1936e40, 0xc0000e01c0, 0xc000579100)
        /usr/local/go/src/net/http/server.go:1995 +0x44
github.com/gorilla/mux.(*Router).ServeHTTP(0xc0000da240, 0x1936e40, 0xc0000e01c0, 0xc000395e00)
        /home/matt/go/pkg/mod/github.com/gorilla/mux@v1.7.0/mux.go:212 +0xe3
net/http.serverHandler.ServeHTTP(0xc0006c2a90, 0x1936e40, 0xc0000e01c0, 0xc000395e00)
        /usr/local/go/src/net/http/server.go:2774 +0xa8
net/http.(*conn).serve(0xc000706a00, 0x1938a00, 0xc000625240)
        /usr/local/go/src/net/http/server.go:1878 +0x851
created by net/http.(*Server).Serve
        /usr/local/go/src/net/http/server.go:2884 +0x2f4
2019/11/26 19:58:17 Web deferred internal error render
2019/11/26 19:58:17 "GET /mention:@matt@write.as" 500 324.540684ms "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/73.0.3683.103 Safari/537.36"
```

qwazix:
This is already fixed in upstream _activityserve_ but that one uses upstream _httpsig_ and not _writeas/httpsig_ (I made those changes as a result of the review of _pherephone_) so I can't update it. Should I make the required changes to writefreely to support upstream _httpsig_?

qwazix:
It turns out this is not necessary. If you do

```
go get github.com/go-fed/httpsig@master
go get -u github.com/writeas/activityserve
```
everything should work.

However the new httpsig removes the burden of setting the digest header manually and signs that header too so it might be a good idea to use that throughout.

thebaer:
Awesome, just verified that that fixes the issue. I'll update `go.mod`, and we can work on moving to `httpsig` in the future.

Did more testing with [this post](). Last issue I'll address:

* Invalid mentions link takes user to a blank page. I'll make it so user gets better feedback ("Profile not found").

For the future, a few notes:

* Funkwhale's response:
```
2019/11/27 14:08:05 Status  : 403 Forbidden
2019/11/27 14:08:05 Response: {"detail":"{'summary': [ErrorDetail(string='This field may not be blank.', code='blank')]}"
```
* Funkwhale mentions link to [raw Actor data](https://open.audio/federation/actors/bear). Not sure if this is something we need to eventually fix on our end, though (e.g. in our case, when web visitors go to the raw actor URL, WF redirects them to the web-based profile).
* PeerTube's response: `403 Forbidden`
* Pherephone mention link works correctly :+1:
* When `POST`ing to the Pherephone inbox, maybe it should also return a `403 Forbidden` like PeerTube and Funkwhale? Not sure.
* microblog.pub seems to work correctly (haven't tested on receiving end), returning a `201 Created`

dariusk:
> Funkwhale mentions link to raw Actor data. Not sure if this is something we need to eventually fix on our end, though 

My two cents: this is an issue on their end, they should redirect based on the request headers if it's a request from a web browser. Especially since they don't provide a link to the human-rendering URL in the raw JSON data.

thebaer:
@dariusk Good point -- just opened an issue for this: [funkwhale#979](https://dev.funkwhale.audio/funkwhale/funkwhale/issues/979).

gytisrepecka:
> * Maybe changing the style of mention links. Initial thoughts are to make the domain part of a user's handle faded out a little (but not completely hidden like on Mastodon). I'm also open to any ideas anyone has here

Wouldn't suggest hiding domain part - it would be the same as hiding domain part in e-mail address. Perhaps only makes sense to show only username when it is on the same instance (domain), but that wouldn't be the case - I believe in most scenarios users will be mentioning Mastodon account while actually posting from WriteFreely blog.

thebaer:
I merged our latest changes, did some cleanup, moved the remote user route to `/@/` instead of `/mention:`, and changed it so the user gets visual feedback if they mention a remote user that doesn't exist.

I think there's still some work to be done around the `datastore.GetProfilePageFromHandle()` func, particularly when it comes to users that don't exist. In short, we need to be able to communicate that information back up through the stack, and really we shouldn't have all that network logic inside a database function. So it still needs a lot of refactoring and logic clean up. But we can address that later. For now, this gets the job done. Merging now!

PR #197, opened 2019-10-15
add basic API endpoint for rendering markdown

robjloranger:
This adds a new `POST` endpoint `/api/generate/markdownify` which takes a JSON body:
```json
{"raw_body": "raw markdown goes here"}
```
and a header `Content-Type: application/json` which will return a JSON response containing the html version.
```json
{"body": "html lies here"}
```
---


robjloranger:
exact endpoint to be discussed

robjloranger:
ready for re-review @thebaer

thebaer:
Great :+1:

Last change: I think we should still return that `out` struct (with the `body` property) that you had before. That way, we get backward-compatibility if we ever want to return additional properties along with the rendered Markdown in the future.

PR #202, opened 2019-10-31
add account deletion

robjloranger:
CLI only but backend supports calls from app.db.DeleteAccount already

takes --delete-account user_id_number with optional --posts to also
delete posts. if --posts is omitted all user posts will be updated to
anonymous posts

[Describe the pull request here...]

---


mrvdb:
Nice! For writefreely hosters, this will be very useful to clean out their 'registered but nothing else'-users after a while.

robjloranger:
After this we will likely be adding a nice button to each user page in the admin view as well

thebaer:
@robjloranger Haven't had a chance to run yet, but after reading through: let's have the `--delete-account` flag take a username instead of the user ID. Requiring the numeric ID isn't too user-friendly and I don't imagine it'll be used much. (Though I'm open to input -- any thoughts, @mrvdb?)

Also, let's have the default behavior delete all user posts -- I'm not sure people will want to leave user posts, especially because future deletion will require manual database work. Or was there a reason you had in mind for keeping posts?

robjloranger:
@thebaer I had the option for posts because it was mentioned in the original task, probably more for write.as though. So for now I can just remove the option to keep posts.

thebaer:
I brought this up to date with `develop` and fixed a small issue I noticed in the admin panel after deleting a user. Otherwise this is good to go!

As mentioned in [T319](https://writefreely.org/tasks/319), this event needs to be federated. So I've created tasks for tackling that: [T720](https://writefreely.org/tasks/720) for posts and [T721](https://writefreely.org/tasks/721) for collections.

PR #203, opened 2019-11-05
T319 admin delete acct

robjloranger:
adds an admin option to delete a user account from the web UI

---

This is built on top of #202 and will need to be rebased after that is in.


thebaer:
We should address federation ([T720](https://writefreely.org/tasks/720) and [T721](https://writefreely.org/tasks/721)) before merging this.

thebaer:
As mentioned on [T319](https://phabricator.write.as/T319#14477), this is a widely-needed feature, so I think it's okay to merge without having the fediverse deletion side in place.

thebaer:
Merging now.

PR #204, opened 2019-11-05
T319 user delete acct

robjloranger:
this adds a user UI component on the settings page to allow account deletion by the user

---

requires #202 and will maybe need a rebase


thebaer:
We should address federation ([T720](https://writefreely.org/tasks/720) and [T721](https://writefreely.org/tasks/721)) and then finish #203 before merging this.

thebaer:
This is ready for review. For admins, two things to note:

You'll want to enable **open account deletions** through the Admin Dashboard or your config file (under the `[app]` section, add: `open_deletion = true`).

This adds CSRF protection on the account deletion endpoint, which requires a new encryption key. To generate that, run:

```
writefreely keys generate
```

thebaer:
Merging now.

PR #205, opened 2019-11-07
Fix URLs in CSV exports

thebaer:
This includes the instance's hostname in calls to export a CSV file and `PublicPost.CanonicalURL()`.

It also fixes a panic in that method during CSV export caused by draft posts.

### DIFF
--- a/posts.go
+++ b/posts.go
@@ -1380,12 +1380,14 @@ Are you sure it was ever here?`,
 			IsFound        bool
 			IsAdmin        bool
 			CanInvite      bool
+			Hostname       string
 		}{
 			PublicPost:     p,
 			StaticPage:     pageForReq(app, r),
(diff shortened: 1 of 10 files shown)


posts.go
```diff
@@ -1380,12 +1380,14 @@ Are you sure it was ever here?`,
 			IsFound        bool
 			IsAdmin        bool
 			CanInvite      bool
+			Hostname       string
```
thebaer: The app configuration is passed into every page, so we don't need this extra property. We can just use `.Host` in the templates.

robjloranger: Oh cool, I can fix it but won't be until later today.

thebaer: Cool, no worries!

robjloranger:
ok all set

thebaer:
Perfect :+1: Thanks for the review! Merging now.

PR #206, opened 2019-11-09
Optimized regexp for matching tags.

emorozov:
A tiny optimization for the SQLite regexp matching removing no-op `.*` before and after the regexp matching tags. I'm not sure if it affects performance noticeably, but it's an inaccurate regexp because `.*` will match anything and doesn't really have to be included.

---

### DIFF
--- a/database.go
+++ b/database.go
@@ -1187,7 +1187,7 @@ func (db *datastore) GetPostsTagged(cfg *config.Config, c *Collection, tag strin
 	var rows *sql.Rows
 	var err error
 	if db.driverName == driverSQLite {
-		rows, err = db.Query("SELECT "+postCols+" FROM posts WHERE collection_id = ? AND LOWER(content) regexp ? "+timeCondition+" ORDER BY created "+order+limitStr, collID, `.*#`+strings.ToLower(tag)+`\b.*`)
+		rows, err = db.Query("SELECT "+postCols+" FROM posts WHERE collection_id = ? AND LOWER(content) regexp ? "+timeCondition+" ORDER BY created "+order+limitStr, collID, `#`+strings.ToLower(tag)+`([,.!\s]|$)`)
 	} else {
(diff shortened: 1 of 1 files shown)


emorozov:
Second commit adds ability to support non-ascii tags. My blog is in Russian and I was all the time wondering why only tags in English work. So I did study writefreely sources and some of the Golang docs, and it seems that Golang regexp fully supports only English language. For example, `\b` works only for English, it won't match word boundary for any word written in non-English alphabet.

I've added a very simple workaround that attempts to match `#tag_text[whitespace|punctuation]` when searching for a tags. While it doesn't solve all the issues with poor handling of non-English languages by Golang and SQLite (e.g. `LOWER(content)` part of the query converts only English text to lower case) but it is good enough for the most cases.

thebaer:
Thanks for contributing this. While I think this is a good stopgap for instances that need it, I'd prefer we implement a more permanent fix that supports all possible character sets on the front and back end. Part of this work will likely involve larger database changes, including tracking hashtags associated with posts instead of doing things with regular expressions.

I'll leave this open for now, but ideally we can fix #219 with a more robust system than what we have today.

emorozov:
Isn't it better to have a solution that is not perfect but works today while waiting for a better system? It's your choice, of course, but this PR replaces one regexp with another regexp, so while it's not a perfect solution, but original regexp isn't optimal either. E.g. `.*` at the beginning and end is just a NOP that wastes regexp engine time.

thebaer:
I agree and am fine with removing the `.*` patterns, but I'm not a fan of introducing new regressions in the word boundary matching just to fix this (e.g. it breaks on `#tag;` or `#tag)`). If we can get SQLite matching closer to the POSIX word boundary we use in the MySQL query, I'll be happy to merge that.

thebaer:
Closing now since there hasn't been any progress. If you want to make those improvements, please feel free to reopen this!

PR #208, opened 2019-11-11
add silenced warning on invites page

robjloranger:
This adds the same warnng about a user account being silence on the view invites page as a silenced use can no longer create invites.

---

### DIFF
--- a/invites.go
+++ b/invites.go
@@ -56,12 +56,20 @@ func handleViewUserInvites(app *App, u *User, w http.ResponseWriter, r *http.Req
 
 	p := struct {
 		*UserPage
-		Invites *[]Invite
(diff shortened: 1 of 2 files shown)


thebaer:
Part of [T661](https://phabricator.write.as/T661).

thebaer:
Since the user can't create invites, I updated the form to also reflect this fact. Now elements are disabled and the submit button includes an explanation tooltip.

Thanks for this! Merging now.

thebaer:
Ah, also page rendering shouldn't catastrophically fail when the `IsUserSuspended` check fails. The 500 will confuse people and the error isn't that critical. Fixed that, merging now.

PR #212, opened 2019-11-19
Fix suspension check in fetchPost()

thebaer:
Previously, this check would return a "user not found" error when retrieving a collection post by its post ID, e.g. `/api/posts/abc123` instead of `/api/collections/demo/posts/my-slug` -- this happens particularly when `Announce`ing a post in the fediverse. This change fixes that.


thebaer:
Thanks @robjloranger! Merging now.

PR #214, opened 2019-11-26
Fix fediverse post IDs

thebaer:
Previously, fetching ActivityStreams data about a post via `/api/posts/ID`, instead of `/api/collections/ALIAS/posts/SLUG` wouldn't include the instance's base URL in the post's `id`, causing federation issues. This fixes that.

It also includes the related fix in #212.


thebaer:
Thanks! Merging now.

PR #215, opened 2019-11-28
Prevent failed requests on failed user silence check

thebaer:
This avoids requests completely failing when a user suspension check fails, such as when `IsUserSuspended()` indicates that a "user isn't found".

There's still some work to do. For example, not-found pages (like `/siteinfo.json`) previously showed:

```
{"error": "Post not found."}
```

Before this fix, they'd show a "Server error" page and a 500.

With this fix, they show:

```
{"error": "Post not found."}<!DOCTYPE HTML>
<html>
	<head>
		<title>Page not found &mdash; WF MySQL Multi?</title>
		...
```

(and erroneously send two status headers).

### DIFF
--- a/posts.go
+++ b/posts.go
@@ -384,7 +384,6 @@ func handleViewPost(app *App, w http.ResponseWriter, r *http.Request) error {
 	suspended, err := app.db.IsUserSuspended(ownerID.Int64)
 	if err != nil {
 		log.Error("view post: %v", err)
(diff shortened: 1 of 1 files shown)


thebaer:
Actually, previously mentioned issue isn't a regression. We'll need to do some refactoring to fix the issue, but that's out of scope here.

Made one more change that prevents an unnecessary log for the user silence check.

PR #218, opened 2019-12-01
Suppress log when editing a collection post or its metadata

### LINKED ISSUE #216: collections.go:233: [PROGRAMMER ERROR] WARNING: Collection.hostName is empty!
```
ERROR: 2019/11/30 09:33:27 collections.go:233: [PROGRAMMER ERROR] WARNING: Collection.hostName is empty! Federation and many other things will fail! If you're seeing this in the wild, please report this bug and let us know what you were doing just before this: https://github.com/writeas/writefreely/issues/new?template=bug_report.md
ERROR: [repeated a few times]
```

Error message popped up upon first follower from a Mastodon instance. Follow worked but posts can't be loaded it seems ("No toots here!").

New single-user writefreely instance, version `v0.11.1`.

thebaer:
This adds the instance's Hostname to the collection data loaded when editing a collection post or its metadata. While not technically needed in this situation, it suppresses an alarming error log.

Resolves #216

### DIFF
--- a/pad.go
+++ b/pad.go
@@ -92,6 +92,7 @@ func handleViewPad(app *App, w http.ResponseWriter, r *http.Request) error {
 		if err != nil {
 			return err
 		}
+		appData.EditCollection.hostName = app.cfg.App.Host
 	} else {
 		// Editing a floating article
 		appData.Post = getRawPost(app, action)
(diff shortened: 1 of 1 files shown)


thebaer:
Thanks for looking at it, @robjloranger. Will merge now.

PR #220, opened 2019-12-05
Support ActivityPub Notes

thebaer:
**This is an experimental change to see what `Note` support in WriteFreely might look like.**

This proof-of-concept federates posts that don't include a double-line break as Notes -- otherwise they go out as Articles, as before.

The idea to use this indicator to separate Notes and Articles was inspired by a [SocialHub discussion](https://socialhub.activitypub.rocks/t/resolving-the-note-vs-article-distinction/258).

Another option for marking a post as a `Note` might be to use a `#note` hashtag. Any other ideas?


thebaer:
If we're getting wild, I'd love to see this be totally configurable -- maybe as an admin-defined value or small script that determines Note vs. Article.

upstreamist:
Similar suggestion here: https://discuss.write.as/t/full-text-when-federating-short-posts/1642 — difference being that posts would go out as Note if they (a) have no user-defined title, and (b) contain 280 characters or fewer. Otherwise, they'd publish as Article.

Character limit could be adjustable, given that some Fediverse platforms allow more that 280 characters.

If you want a more user-deliberate way to flag different content types, I'd suggest adapting the @-name format, rather than using a hashtag — i.e. address a new post to @note to send it as a Note. Likewise, @Article could specify Article-type, but should be superfluous in most cases, assuming Article is the default type.

If @-name causes too much confusion, or conflicts with potential Webmention implementation, maybe a less familiar notation, like &note, would work better.

Whatever the final decision, one setting I'd like to see implemented with this is the option to only federate Note-type posts, or federate both Note and Article.

thebaer:
Merging now!

PR #230, opened 2019-12-30
OAuth Provider: Slack

ngerakines:
This PR introduces Slack specific functionality as part of the larger OAuth feature set.

This PR builds off of the functionality in https://github.com/writeas/writefreely/pull/225.

## Testing

1. Create a Slack application
2. Set the following configuration

       [oauth.slack]
       client_id = "<slack application client id>"
       client_secret = "<slack application client secret>"

3. Browse to https://[local instance]/oauth/slack
4. Continue authentication flow with Slack
5. Arrive at the logged in user landing page with your newly created account.

*Important*: Be sure to set your instance to allow multiple users in configuration.

---

### DIFF
--- a/oauth.go
+++ b/oauth.go
@@ -34,6 +34,7 @@ type InspectResponse struct {
 	ExpiresAt time.Time `json:"expires_at"`
 	Username  string    `json:"username"`
 	Email     string    `json:"email"`
+	Error     string    `json:"error"`
 }
 
 // tokenRequestMaxLen is the most bytes that we'll read from the /oauth/token
@@ -104,7 +105,7 @@ func configureSlackOauth(r *mux.Router, app *App) {
 			ClientSecret:     app.Config().SlackOauth.ClientSecret,
 			TeamID:           app.Config().SlackOauth.TeamID,
 			CallbackLocation: app.Config().App.Host + "/oauth/callback",
-			HttpClient:       &http.Client{Timeout: 10 * time.Second},
+			HttpClient:       config.DefaultHTTPClient(),
 		}
 		configureOauthRoutes(r, app, oauthClient)
 	}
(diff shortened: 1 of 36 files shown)


oauth.go
```diff
@@ -56,54 +55,85 @@ type OAuthDatastoreProvider interface {
...
 
-	return u.String(), nil
+func configureWriteAsOauth(r *mux.Router, app *App) {
+	if app.Config().WriteAsOauth.ClientID != "" {
+		oauthClient := writeAsOauthClient{
+			ClientID:         app.Config().WriteAsOauth.ClientID,
+			ClientSecret:     app.Config().WriteAsOauth.ClientSecret,
+			ExchangeLocation: app.Config().WriteAsOauth.TokenLocation,
+			InspectLocation:  app.Config().WriteAsOauth.InspectLocation,
+			AuthLocation:     app.Config().WriteAsOauth.AuthLocation,
```
thebaer: I think we should fall back to the default Write.as endpoints here if config values are empty. Thoughts?

oauth_slack.go
```diff
@@ -0,0 +1,148 @@
...
+		return nil, err
+	}
+	return inspectResponse.InspectResponse(), nil
+}
+
+func (resp slackUserIdentityResponse) InspectResponse() *InspectResponse {
+	return &InspectResponse{
+		UserID:   resp.User.ID,
+		Username: slug.Make(resp.User.Name),
+		Email:    resp.User.Email,
```
thebaer: If we could also feed the original `resp.User.Name` into the user creation process, so it becomes the default blog title (like what happens in `datastore.CreateUser()`), that would be awesome.

oauth.go
ngerakines: Yep, good call.

oauth.go
```diff
@@ -152,116 +181,56 @@ func (h oauthHandler) viewOauthCallback(w http.ResponseWriter, r *http.Request)
...
 			HashedPass: hashedPass,
 			HasPass:    true,
-			Email:      zero.NewString("", tokenInfo.Email != ""),
+			Email:      zero.NewString(tokenInfo.Email, tokenInfo.Email != ""),
 			Created:    time.Now().Truncate(time.Second).UTC(),
 		}
+		displayName := tokenInfo.DisplayName
+		if len(displayName) == 0 {
+			displayName = tokenInfo.Username
+		}
```
thebaer: We also need to encrypt the user's email address, as is done here: https://github.com/writeas/writefreely/blob/f343cebce75794660538952d830fe84a0e2e8f8c/account.go#L162-L169

oauth.go
```diff
@@ -152,116 +181,56 @@ func (h oauthHandler) viewOauthCallback(w http.ResponseWriter, r *http.Request)
...
 
 		if err := loginOrFail(h.Store, w, r, newUser); err != nil {
-			failOAuthRequest(w, http.StatusInternalServerError, err.Error())
+			return impart.HTTPError{http.StatusInternalServerError, err.Error()}
 		}
-		return
+		return nil
 	}
 
 	user, err := h.DB.GetUserForAuthByID(localUserID)
```
thebaer: This database func doesn't fetch the user's `username` from the database, so on subsequent logins the user doesn't have a `Username` set, which causes funkiness in the UI.

After looking through the code, it looks like `GetUserForAuthByID()` is functioning as intended, but is just poorly named (see usage in the `ChangeSettings()` func). At any rate, calling `GetUserByID()` instead should fix this.

PR #231, opened 2019-12-30
Pass OAuth requests through OAuth handler

thebaer:
This gives us consistent logging on OAuth endpoints and passes around errors with `impart.HTTPError`, as we do elsewhere -- while keeping the current JSON schema for OAuth requests (`{"error": "..."}`).

@ngerakines curious about your thoughts on this -- whether it works or if it might cause issues down the road, e.g. with error handling for various request types.

Part of [T705](https://writefreely.org/tasks/705).


thebaer:
Looks great! Merging now.

PR #232, opened 2020-01-03
OAuth registration improvements

ngerakines:
This PR improves the account creation flow through OAuth providers. Instead of immediately creating a user, a populated form is presented with suggested values. Some basic error handling is implemented.

### Added

* The /oauth/signup handler, template, and route.
* The `server.hash_seed` configuration variable used to verify the integrity of oauth variables used in account creation.

### Changed

* The account creation behavior used when a user account should be created as a result of the oauth flow.

---


oauth_signup.go
```diff
@@ -0,0 +1,209 @@
...
+	}
+	alias := r.FormValue(oauthParamAlias)
+	if len(alias) < 5 {
+		return impart.HTTPError{Status: http.StatusBadRequest, Message: "Alias is too short."}
+	}
+	if len(alias) > 20 {
+		return impart.HTTPError{Status: http.StatusBadRequest, Message: "Alias is too long."}
+	}
+	password := r.FormValue("password")
+	if len(password) < 5 {
```
thebaer: I think password constraints like this would be best in a configuration value (like `MinUsernameLen`) so admins can choose whether or not to introduce these. Probably best to leave this to another task -- just created [T716](https://writefreely.org/tasks/716) for this.

pages/signup-oauth.tmpl
```diff
@@ -0,0 +1,38 @@
...
+	    <input type="hidden" name="token_email" value="{{ .TokenEmail }}" />
+	    <input type="hidden" name="token_remote_user" value="{{ .TokenRemoteUser }}" />
+	    <input type="hidden" name="provider" value="{{ .Provider }}" />
+	    <input type="hidden" name="client_id" value="{{ .ClientID }}" />
+	    <input type="hidden" name="signature" value="{{ .TokenHash }}" />
+
+		<input type="text" name="username" placeholder="Username" value="{{.Username}}" /><br />
+		<input type="text" name="alias" placeholder="Alias"{{ if .Alias }} value="{{.Alias}}"{{ end }} /><br />
+		<input type="text" name="email" placeholder="Email"{{ if .Email }} value="{{.Email}}"{{ end }} /><br />
+		<input type="password" name="password" placeholder="Password" /><br />
```
thebaer: We should keep the `<label>`s from signup.tmpl on the fields here, so people know what each one is for.

oauth.go
```diff
@@ -99,39 +117,57 @@ func (h oauthHandler) viewOauthInit(app *App, w http.ResponseWriter, r *http.Req
...
-			ClientSecret:     app.Config().SlackOauth.ClientSecret,
-			TeamID:           app.Config().SlackOauth.TeamID,
-			CallbackLocation: app.Config().App.Host + "/oauth/callback",
-			HttpClient:       config.DefaultHTTPClient(),
+			ClientID:     app.Config().SlackOauth.ClientID,
+			ClientSecret: app.Config().SlackOauth.ClientSecret,
+			TeamID:       app.Config().SlackOauth.TeamID,
+			HttpClient:   config.DefaultHTTPClient(),
+			//CallbackLocation: app.Config().App.Host + "/oauth/callback",
+			CallbackLocation: "http://localhost:5000/callback",
```
thebaer: Can we make this endpoint configurable somehow? Maybe in `SlackOauth`, and falling back to the local endpoint if there's no config value?

oauth.go
```diff
@@ -99,39 +117,57 @@ func (h oauthHandler) viewOauthInit(app *App, w http.ResponseWriter, r *http.Req
...
+		}
 		oauthClient := writeAsOauthClient{
 			ClientID:         app.Config().WriteAsOauth.ClientID,
 			ClientSecret:     app.Config().WriteAsOauth.ClientSecret,
 			ExchangeLocation: config.OrDefaultString(app.Config().WriteAsOauth.TokenLocation, writeAsExchangeLocation),
 			InspectLocation:  config.OrDefaultString(app.Config().WriteAsOauth.InspectLocation, writeAsIdentityLocation),
 			AuthLocation:     config.OrDefaultString(app.Config().WriteAsOauth.AuthLocation, writeAsAuthLocation),
 			HttpClient:       config.DefaultHTTPClient(),
-			CallbackLocation: app.Config().App.Host + "/oauth/callback",
+			CallbackLocation: "http://localhost:5000/callback",
```
thebaer: Ditto above comment here

PR #233, opened 2020-01-05
Rename base_url to collection_url in Markdown API

thebaer:
This renames the `base_url` parameter to `collection_url`, to make it clear what the parameter is for (i.e. not the base URL of the instance, but of the collection itself).

Part of [T519](https://writefreely.org/tasks/519)


thebaer:
Cool, thanks for looking at it :+1:

PR #235, opened 2020-01-09
Add dates to blog posts

thebaer:
This shows the publish date for collection posts when a collection is set to the "Blog" display format, making individual post pages consistent with the index page.

This resolves [T669](https://writefreely.org/tasks/669).


robjloranger:
I haven't tested but I'm happy with this, especially if it matches the current style better.

thebaer:
Cool, thanks! Merging now.

PR #236, opened 2020-01-10
Fixing bug in oauth callback URL registration.

ngerakines:
Fixing a bug in the oauth callback URL registration where the lack of provider context was overwriting the previous oauth callback route registration call.

---


thebaer:
Looks good to me, thanks!

PR #239, opened 2020-01-14
Switch to a maintained fork of XGO

techknowlogick:
👋 Hey, I'm one of the project leads for [Gitea](https://gitea.io/en-us/), and we also use XGO to cross compile our binaries, however we have forked xgo to ensure that versions are kept up to date and have added other enhancements.

I've also added arm-6 as a compile target, this is due to its wider support on arm based devices (arm6 works on arm7 devices, but not the reverse).

Note: Due to switching from Ubuntu 14.04 to 18.04 as a base image in the fork this prevents binaries from working on Linux Kernel versions below 3.1. However due to Ubuntu 14.04 being past EOL, and also that Linux Kernel version being of a certain age Gitea proceeded with that being acceptable (it may not be acceptable for you).

I understand that this is an unsolicited PR, and it changes an upstream dependency, as such please feel free to close it if it doesn't align with your project. 

---

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,6 +41,12 @@ build-darwin: deps
 	fi
 	xgo --targets=darwin/amd64, -dest build/ $(LDFLAGS) -tags='sqlite' -out writefreely ./cmd/writefreely
 
+build-arm6: deps
+	@hash xgo > /dev/null 2>&1; if [ $$? -ne 0 ]; then \
+		$(GOGET) -u src.techknowlogick.com/xgo; \
+	fi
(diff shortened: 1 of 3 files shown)


thebaer:
Hey @techknowlogick, thank you for contributing this! We'd be happy to accept it.

Everything looks good, though running it modifies `go.mod` and `go.sum` -- would you mind committing those changes? Then we can merge this in.

techknowlogick:
@thebaer just added the modifications to go modules, please let me know if I can make any other changes :)

thebaer:
Looks great, thanks! Merging now.

PR #241, opened 2020-01-14
fixes imported post times

robjloranger:
changes the client side to round the unix time to avoid floats

alters the time to match the client time zone on the server side

---

### DIFF
--- a/account_import.go
+++ b/account_import.go
@@ -9,6 +9,7 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


account_import.go
```diff
@@ -147,6 +149,12 @@ func handleImport(app *App, u *User, w http.ResponseWriter, r *http.Request) err
 			post.Collection = collAlias
 		}
 		dateTime := time.Unix(fileDates[formFile.Filename], 0)
+		offset, err := strconv.Atoi(fileTZ)
+		if err != nil {
+			log.Error("form time zone offset not a valid integer: %v", err)
+			continue
+		}
+		dateTime = dateTime.Add(time.Minute * time.Duration(offset))
```
thebaer: I think, design-wise, ideally we'd do all of this on the client-side, just like we do in `templates/edit-meta.tmpl`. That keeps the API clean, consistent, and the time logic all in one place, so it's always easier to debug.

thebaer:
Looks great, thanks!

PR #243, opened 2020-01-15
OAuth account management

ngerakines:
This PR introduces oauth account management functionality on the account settings page. With it, oauth accounts can be added and removed from existing accounts.

---


migrations/v6.go
```diff
@@ -0,0 +1,36 @@
...
+	}
+	return wf_db.RunTransactionWithOptions(context.Background(), db.DB, &sql.TxOptions{}, func(ctx context.Context, tx *sql.Tx) error {
+		builders := []wf_db.SQLBuilder{
+			dialect.
+				AlterTable("oauth_client_states").
+				AddColumn(dialect.
+					Column(
+						"attach_user_id",
+						wf_db.ColumnTypeInteger,
+						wf_db.OptionalInt{Set: true, Value: 24,}).SetNullable(false).SetDefault("0")),
```
thebaer: Should this column be NULLable for the cases where we're not attaching a user?

ngerakines: Personal preference, but I can change it. I avoid null like the plague. I'd rather have default values for primatives whenever possible. I think it translates to Go better and reduces complexity of simple values because you don't have to do "is this is set *and/or* is this a non-default value?" validation.

thebaer: Yeah, it's definitely nicer in Go without them -- they're a pain to work with. But in general, I think it's good database design to put NULLs in optional columns. It conveys the right information in the data itself (e.g. this column was designed to be empty sometimes vs. maybe a bug made this column empty), and helps if we ever need to troubleshoot or to change it somehow. And overall for the database design, I think consistency is best.

Of course, I also think it's fine to just return an `int64` (instead of `sql.NullInt64`) from our database funcs that fetch the value, as you're already doing.

thebaer:
I went ahead and updated the schema and logic around that column so it's a `NULL` now. Only issue I ran into was this:

1. Already have a WF _Account A_ connected to a Slack account
1. I log out of it
1. I log into another WF _Account B_, then choose to attach Slack
1. I go through the normal flow to attach the Slack account to _Account B_, but it brings me back to WF logged in as _Account A_.

Ideally, we could show an error in this case or some other kind indication to the user, as this is a little confusing.

thebaer:
Looks great! Verified the issue is fixed.

I tweaked the design and wording on the Account Settings page (see below), and made a few other changes. Now I'm just verifying different configurations and edge cases. If that looks good, I'll merge.

### No accounts linked

Screenshot from 2020-03-24 10-33-13

### Slack linked

Screenshot from 2020-03-24 10-30-40

thebaer:
Merging now! Thanks @ngerakines!

PR #244, opened 2020-01-16
OAuth signup form tweaks

thebaer:
This fixes / adds a few things during OAuth signup (see commits for more notes):

* Reserve the username "oauth"
* Generate non-colliding usernames in all lowercase
* Fix false login state on OAuth signup page
* Add comments for future refactoring

Including changes designed to reduce friction during the signup process: 

* Add async username check on OAuth signup form
* Fall back to the given username as the "Display Name" (collection title) if none given during signup
* Remove Password field / requirement from OAuth signup page, to eliminate a bit of friction. Resolves [T715](https://phabricator.write.as/T715)

### DIFF
--- a/author/author.go
+++ b/author/author.go
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2018 A Bunch Tell LLC.
+ * Copyright © 2018-2020 A Bunch Tell LLC.
  *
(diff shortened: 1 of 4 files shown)


thebaer:
Thanks!

PR #245, opened 2020-01-16
Require authenticated user for editor access

thebaer:
Previously, anyone could access the editor at `/new` (multi-user) or `/me/new` (single-user) even if they weren't logged in. They couldn't do much in that case (publishing would fail), but it could potentially cause some confusion.

Now, users will be sent to the login page, and then redirected back to the editor once successfully logged in.


thebaer:
Thanks! Merging.

PR #247, opened 2020-01-20
update upgrade script for recent changes

robjloranger:
changes accounted for
- the tar directory structure had changed to use a subdirectory
- there are now multiple linux targets released

bugs
- the service must be stopped before replacing the binary
- migrations were not being run during an upgrade

---


thebaer:
Thanks for making these fixes! LGTM. I just changed "Restarting..." to "Starting..." to match the new action.

Merging now.

PR #249, opened 2020-01-23
Improve resource use for federation

thebaer:
This is a work in progress that'll include various improvements to federation with ActivityPub. Included:

* Add timeout on ActivityPub requests. Previously, if a fediverse instance was slow or offline, requests could hang indefinitely. This should fix that.


thebaer:
Verified to work well on Write.as over the past couple weeks. Merging now.

PR #254, opened 2020-01-29
Fix date format in `datetime` attribute

### LINKED ISSUE #253: All posts display 'Invalid Date'
I just pulled `develop` and now all posts will display an 'Invalid Date' just below the title

I'm guessing (but have not tested) this is related to the latest 'local date' munging I saw flying by.

The error occurs in `static/js/localdate.js`:

`var d = new Date(dateEl.getAttribute("datetime"));`

value of the attribute was: `2019-03-05 13:12:54 +0100 CET` which indeed gives an `Invalid date` if passed into the js `Date()` constructor

Anything I missed?

thebaer:
Previously, the date format in this attribute for posts was invalid. This caused local date rendering to fail in Firefox. This fixes that.

Closes #253

Also in this PR: fixed localdate.js not getting included on the tags page when a user was unauthenticated.

### DIFF
--- a/templates/chorus-collection-post.tmpl
+++ b/templates/chorus-collection-post.tmpl
@@ -58,7 +58,7 @@ body#post header {
 		{{if .Suspended}}
 			{{template "user-suspended"}}
 		{{end}}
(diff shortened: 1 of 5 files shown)


mrvdb:
I think you missed one (or two) in template read.tmpl

thebaer:
Good catch, thanks! Just fixed.

PR #261, opened 2020-02-10
Refactor CLI

techknowlogick:
:wave: Hey, I know this is kinda a big PR and I'm thankful for you taking the time to even consider it. First, I want to say I understand that if this doesn't align with the current goal of the project that it may be closed, and that's totally ok with me.

Currently the CLI uses flags for both commands and options, this refactor switches it out to a format that allows for bashcompletion built in, and can auto-generate manpages.

For legacy sake, previous style of working with CLI still exists.

If the previous style of using only flags for everything is preferable, then this PR can be changed to still use `urfave/cli`, but still be flag only and would still provide the benefits of the refactored code style.

Please let me know if you have any questions.

Example help output:
```
NAME:
   WriteFreely - A beautifully pared-down blogging platform

USAGE:
   writefreely [global options] command [command options] [arguments...]

VERSION:
   WriteFreely 0.11.2

COMMANDS:
   user        user management tools
   db          db management tools
   config      config management tools
   keys        key management tools
   serve, web  Run web application
   help, h     Shows a list of commands or help for one command

GLOBAL OPTIONS:
   -c FILE        Load configuration from FILE (default: "config.ini")
   --debug        Enables debug logging (default: false)
   --help, -h     show help (default: false)
   --version, -v  print the version (default: false)
```
---

### DIFF
--- a/cmd/writefreely/config.go
+++ b/cmd/writefreely/config.go
@@ -30,25 +30,23 @@ var (
 		Name:    "generate",
 		Aliases: []string{"gen"},
 		Usage:   "Generate a basic configuration",
-		Action: genConfigAction,
+		Action:  genConfigAction,
 	}
 
 	cmdConfigInteractive cli.Command = cli.Command{
-		Name:    "interactive",
-		Aliases: []string{"i"},
-		Usage:   "Interactive configuration process",
+		Name:   "start",
+		Usage:  "Interactive configuration process",
 		Action: interactiveConfigAction,
 		Flags: []cli.Flag{
 			&cli.StringFlag{
 				Name:  "sections",
 				Value: "server db app",
-				Usage: "Which sections of the configuration to go through (requires --config)\n" +
-					"valid values are any combination of 'server', 'db' and 'app' \n" +
-					"example: writefreely --config --sections \"db app\"",
+				Usage: "Which sections of the configuration to go through\n" +
+					"valid values of sections flag are any combination of 'server', 'db' and 'app' \n" +
+					"example: writefreely config start --sections \"db app\"",
 			},
 		},
 	}
-
 )
 
 func genConfigAction(c *cli.Context) error {
--- a/cmd/writefreely/config.go
+++ b/cmd/writefreely/config.go
@@ -1,0 +1,63 @@
+/*
+ * Copyright © 2020 A Bunch Tell LLC.
+ *
+ * This file is part of WriteFreely.
+ *
+ * WriteFreely is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, included
+ * in the LICENSE file in this source code package.
+ */
+
+package main
+
+import (
+	"github.com/writeas/writefreely"
+
+	"github.com/urfave/cli/v2"
+)
+
+var (
+	cmdConfig cli.Command = cli.Command{
+		Name:  "config",
+		Usage: "config management tools",
+		Subcommands: []*cli.Command{
+			&cmdConfigGenerate,
+			&cmdConfigInteractive,
+		},
+	}
+
+	cmdConfigGenerate cli.Command = cli.Command{
+		Name:    "generate",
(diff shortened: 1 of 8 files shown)


thebaer:
Thanks for working on this! It'll still be a while before I can fully test / review, but wanted to mention that I am definitely interested in merging this.

cmd/writefreely/user.go
```diff
@@ -0,0 +1,89 @@
...
+		Subcommands: []*cli.Command{
+			&cmdAddUser,
+			&cmdDelUser,
+			&cmdResetPass,
+			// TODO: possibly add a user list command
+		},
+	}
+
+	cmdAddUser cli.Command = cli.Command{
+		Name:    "add",
```
thebaer: I think this should be "create" to echo the language we had before. (We can keep "add" as an alias.)

cmd/writefreely/config.go
```diff
@@ -0,0 +1,63 @@
...
+
+	cmdConfigGenerate cli.Command = cli.Command{
+		Name:    "generate",
+		Aliases: []string{"gen"},
+		Usage:   "Generate a basic configuration",
+		Action: genConfigAction,
+	}
+
+	cmdConfigInteractive cli.Command = cli.Command{
+		Name:    "interactive",
```
thebaer: I'd like this to be something memorable and approachable like "start", since most people will be typing this command when they first set up WF.

cmd/writefreely/config.go
```diff
@@ -0,0 +1,63 @@
...
+		Aliases: []string{"i"},
+		Usage:   "Interactive configuration process",
+		Action: interactiveConfigAction,
+		Flags: []cli.Flag{
+			&cli.StringFlag{
+				Name:  "sections",
+				Value: "server db app",
+				Usage: "Which sections of the configuration to go through (requires --config)\n" +
```
thebaer: We can remove "(requires --config)" since this flag is part of the `config` subcommand.

cmd/writefreely/config.go
```diff
@@ -0,0 +1,63 @@
...
+					"valid values are any combination of 'server', 'db' and 'app' \n" +
+					"example: writefreely --config --sections \"db app\"",
```
thebaer: Should update this example to reflect subcommands.

cmd/writefreely/user.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+	app := writefreely.NewApp(c.String("c"))
+	return writefreely.DoDeleteAccount(app, username)
+}
+
+func resetPassAction(c *cli.Context) error {
+	username := ""
+	if c.NArg() > 0 {
+		username = c.Args().Get(0)
+	}
```
thebaer: `if username == ""` after this, we should output usage information and exit. (Otherwise right now this prints out a cryptic error: `app.go:663: Get user: User doesn't exist.`)

cmd/writefreely/main.go
```diff
@@ -11,122 +11,154 @@
...
-			log.Error(err.Error())
-			os.Exit(1)
-		}
-		os.Exit(0)
-	} else if *deleteUsername != "" {
-		err := writefreely.DoDeleteAccount(app, *deleteUsername)
+	app := &cli.App{
+		Name:    "WriteFreely",
+		Usage:   "A beautifully pared-down blogging platform",
+		Version: writefreely.FormatVersion(),
```
thebaer: Running `writefreely -v` now outputs:

```
WriteFreely version WriteFreely 0.11.2-172-gb0f0de3
```

Whereas before it'd be something like:

```
WriteFreely 0.11.2-177-gf799260
```

Ideally we could keep the output format identical to what it was before (removing the "WriteFreely version" prefix).

techknowlogick:
@thebaer thanks for feedback. I've just pushed requested changes :)

thebaer:
Thanks for the quick fixes! Looks great :+1: merging now.

PR #262, opened 2020-02-10
Consistent backend page design

thebaer:
This fixes various pages that used `<h2>`s instead of `<h1>`s:

* Drafts
* Blogs
* Account Settings
* Export

It also fixes the width of content in the Reader page so it's consistent with all other pages.


thebaer:
Pretty straightforward; merging now.

PR #263, opened 2020-02-14
Typography Improvements, Vol. I

thebaer:
This PR aims to introduce small, incremental improvements to the typography on WriteFreely. As these changes will affect all aspects of WF, including user blogs, the goal is to not introduce any major, jarring changes that will negatively impact users.

All typographical input is welcome here -- please feel free to comment with any improvements you would like to see so we can discuss and possibly include them here.

## Changes

* Change line height from 1.4 to 1.5

### DIFF
--- a/less/core.less
+++ b/less/core.less
@@ -644,7 +644,7 @@ body#collection article, body#subpage article {
 	padding-bottom: 0;
 	.book {
 		h2 {
-			font-size: 1.5;
+			font-size: 1.4em;
 		}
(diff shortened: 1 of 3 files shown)


cjeller1592:
Here's a before/after for reference to these typography improvements...

## Before 

Screen Shot 2020-02-26 at 3 01 22 PM

## After 

Screen Shot 2020-02-26 at 3 00 55 PM

thebaer:
Merging now.

PR #267, opened 2020-02-22
Update Contributing guide

thebaer:
This updates the guide to explain our current practices and processes. It now focuses primarily on contributing code, instead of other types of contributions. And it gives more details on the _ideal_ code and git practices we like to see.

[**Preview the full guide here**](https://github.com/writeas/writefreely/blob/update-contributing-guide/CONTRIBUTING.md).

**Is anything still unclear? Any other information we're missing?**


thebaer:
Merging now.

PR #268, opened 2020-02-23
Add 'X-Robots-Tag: noindex' header to invite URLs

thebaer:
This instructs search engines to not index invite links.


thebaer:
Merging now.

PR #269, opened 2020-02-25
Adjust Reader limits and filtering

thebaer:
This adjusts how the Reader displays Public posts to better suit various WriteFreely communities.

### Limit Reader posts by count, not publish date

_From this feature request:_ [Let the Reader display posts older than 3 months](https://discuss.write.as/t/let-the-reader-display-posts-older-than-3-months/1119).

This changes the Reader to show the 250 most recent posts (keeping the 5-post-per-author limit) instead of only posts from the last 3 months. The previous 3-month limit on posts was a remnant from [Read Write.as](https://read.write.as), where there's a large community of consistent posters. It doesn't make as much sense on most WriteFreely communities.

### DIFF
--- a/read.go
+++ b/read.go
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2018-2019 A Bunch Tell LLC.
+ * Copyright © 2018-2020 A Bunch Tell LLC.
  *
  * This file is part of WriteFreely.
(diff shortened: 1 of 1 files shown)


thebaer:
Merging now.

PR #270, opened 2020-02-26
add js, jsx and html aliases for highlightjs

nkoehring:
This adds aliases for highlightjs that allow to use the pretty common js, jsx abbreviations for javascript and html for xml.

---


cjeller1592:
Looks like it's working for me! Here's a post using the different aliases:

thebaer:
Thanks for contributing this @nkoehring, this is a nice addition. And thanks for reviewing this, @cjeller1592! Looks good to me as well. Merging now.

PR #271, opened 2020-02-26
Make highlightjs theme configurable

nkoehring:
Building on top of #31 this adds an option to the config.ini that allows to set a global theme.

*pros*
 * Allows instances to change the highlightjs theme to their wishes.
 * Works fine for single user instances.

*cons*
 * Admins need to download the themes first. Maybe use a CDN URL instead of a local one?
 * Setting is global for all users.

---

### DIFF
--- a/config.ini.example
+++ b/config.ini.example
@@ -14,6 +14,7 @@ port     = 3306
 site_name         = WriteFreely Example Blog!
 host              = http://localhost:8080
 theme             = write
+highlighter_theme = atom-one-light
 disable_js        = false
 webfonts          = true
 single_user       = true
--- a/config/config.go
+++ b/config/config.go
@@ -63,13 +63,14 @@ type (
 		Host     string `ini:"host"`
 
 		// Site appearance
-		Theme      string `ini:"theme"`
-		Editor     string `ini:"editor"`
-		JSDisabled bool   `ini:"disable_js"`
-		WebFonts   bool   `ini:"webfonts"`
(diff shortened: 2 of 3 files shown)


thebaer:
Thanks for submitting this, @nkoehring! And sorry for the delayed response.

I think it'd be great to allow customization here, but it is tough for multi-user instances -- ideally we'd have something that each user can set themselves, just like they can modify their own CSS.

There's a backend mechanism for handling per-user settings right now that we could use: the `userattributes` database table. But we'd need to add the user interface for updating their preferred style. And ideally, we'd incorporate these styles locally -- a few options:

* By default, include several other themes to choose from locally, or
* Provide a way to automatically download and install themes

nkoehring:
Yes, as mentioned I see the problematic that it only works on a global level. I proposed it as a "better than nothing" version and actually patched my own running instance with it. To provide it on a per user level would be much better obviously.

As a good middle way, and because themes are simply CSS, how about using a CDN and simply download them on the fly by inserting a <link> tag?

thebaer:
Yep, that'd be great if we could download new themes on the fly.

thebaer:
With some more thought, I think the best way to do this would be to include it in the `collectionattributes` database table. That way, whether you have a single user or multi user instance, everything works correctly.

Since this will take a little more time, and potentially a different path to get there, I think we should avoid adding config options that might disappear at some point and close this for now. If you'd like to submit a permanent solution at some point that works across single- and multi-user instances, please feel free!

PR #273, opened 2020-03-02
Add Gopher support

thebaer:
This adds Gopher support to WriteFreely, thanks to @prologic's excellent [go-gopher](https://github.com/prologic/go-gopher) library. (This feature is tracked here: [T559](https://writefreely.org/tasks/559).)

This works on both single- and multi-user instances. It is off by default, but can be enabled with the new `gopher_port` config value in the `[server]` section -- set it to the port you want to run the gopher server on. Then you'll want to expose that port to the internet or proxy connections to it.

When Gopher support is enabled, multi-user instances will show all public blogs at `gopher://[host]:[gopher_port]/` -- otherwise, blogs are accessible at `gopher://[host]:[gopher_port]/[blog]/`

On single-user instances, you can access the blog via `gopher://[host]:[gopher_port]/` and posts via `gopher://[host]:[gopher_port]/[slug]`

~~**This is just a proof of concept for now**. We still need to handle some edge cases and different configurations, like `private = true`. But this should be a good start.~~ This is handled.

What do you think? Any improvements you'd suggest?


prologic:
> This adds Gopher support to WriteFreely, thanks to @prologic's excellent go-gopher library. (This feature is tracked here: T559.)

😲 seriously? Something I wrote is getting serious use? :) WOW!

thebaer:
@prologic Yep! Thanks for the great work :slightly_smiling_face: 

Merging this in now.

PR #274, opened 2020-03-02
Private instance cleanup

thebaer:
This improves things a bit on private instances.

* Hide Reader link in footer when unauthenticated on private instances.
This isn't publicly visible, so as in the header, we now also hide this in the footer.
* Show **Log in** page as landing on private instances, instead of normal landing page.
Private instances are only meant for registered users, and are often closed or invite-only, making the current landing page kind of pointless (and potentially confusing). This remedies that by putting the login form first, while of course still linking to the signup page if an instance has open registrations.

### DIFF
--- a/app.go
+++ b/app.go
@@ -221,6 +221,10 @@ func handleViewHome(app *App, w http.ResponseWriter, r *http.Request) error {
 			return handleViewPad(app, w, r)
 		}
 
+		if app.cfg.App.Private {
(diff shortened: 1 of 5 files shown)


cjeller1592:
Looks good! Really like having the "Reader" link removed in the footer and "Log in" as the default landing page. Definitely clears things up for private instances.

thebaer:
Merging now!

PR #275, opened 2020-03-03
Fix V4 + V5 SQLite migrations

thebaer:
This fixes the unreleased V4 and V5 database migrations so they work with instances backed by SQLite databases.


thebaer:
Thanks for reviewing!

PR #276, opened 2020-03-09
Don't show date on pinned post page

thebaer:
After work from #235, dates showed up on all post pages. However, they don't really fit on pinned posts, which are meant to be timeless. So this fixes that by hiding dates on pinned posts.

Part of [T669](https://writefreely.org/task/669).


thebaer:
Merging now.

PR #277, opened 2020-03-11
Add Gitlab OAuth

paddatrapper:
Adds GitLab support to the OAuth authentication structure.

It is discussed [here](https://discuss.write.as/t/gitlab-oauth/). Currently clicking on the 'Login with GitLab' button results in a 404 error.

---

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -72,9 +72,8 @@ type (
 	GitlabOauthCfg struct {
 		ClientID         string `ini:"client_id"`
 		ClientSecret     string `ini:"client_secret"`
-		AuthLocation     string `ini:"auth_location"`
-		TokenLocation    string `ini:"token_location"`
-		InspectLocation  string `ini:"inspect_location"`
+		Host             string `ini:"host"`
+		DisplayName      string `ini:"display_name"`
 		CallbackProxy    string `ini:"callback_proxy"`
 		CallbackProxyAPI string `ini:"callback_proxy_api"`
 	}
--- a/config/config.go
+++ b/config/config.go
@@ -69,6 +69,16 @@ type (
 		CallbackProxyAPI string `ini:"callback_proxy_api"`
 	}
 
+	GitlabOauthCfg struct {
+		ClientID         string `ini:"client_id"`
+		ClientSecret     string `ini:"client_secret"`
+		AuthLocation     string `ini:"auth_location"`
+		TokenLocation    string `ini:"token_location"`
+		InspectLocation  string `ini:"inspect_location"`
+		CallbackProxy    string `ini:"callback_proxy"`
+		CallbackProxyAPI string `ini:"callback_proxy_api"`
+	}
+
 	SlackOauthCfg struct {
 		ClientID         string `ini:"client_id"`
 		ClientSecret     string `ini:"client_secret"`
@@ -128,6 +138,7 @@ type (
 		App          AppCfg          `ini:"app"`
 		SlackOauth   SlackOauthCfg   `ini:"oauth.slack"`
 		WriteAsOauth WriteAsOauthCfg `ini:"oauth.writeas"`
+		GitlabOauth  GitlabOauthCfg  `ini:"oauth.gitlab"`
 	}
 )
 
--- a/oauth.go
+++ b/oauth.go
@@ -179,12 +179,13 @@ func configureGitlabOauth(parentHandler *Handler, r *mux.Router, app *App) {
 			callbackLocation = app.Config().GitlabOauth.CallbackProxy
 		}
 
+        address := config.OrDefaultString(app.Config().GitlabOauth.Host, gitlabHost)
 		oauthClient := gitlabOauthClient{
 			ClientID:         app.Config().GitlabOauth.ClientID,
 			ClientSecret:     app.Config().GitlabOauth.ClientSecret,
-			ExchangeLocation: config.OrDefaultString(app.Config().GitlabOauth.TokenLocation, gitlabExchangeLocation),
-			InspectLocation:  config.OrDefaultString(app.Config().GitlabOauth.InspectLocation, gitlabIdentityLocation),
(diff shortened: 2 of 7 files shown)


go.mod
```diff
@@ -21,6 +21,7 @@ require (
 	github.com/guregu/null v3.4.0+incompatible
 	github.com/hashicorp/go-multierror v1.0.0
 	github.com/ikeikeikeike/go-sitemap-generator/v2 v2.0.2
+	github.com/jteeuwen/go-bindata v3.0.7+incompatible // indirect
```
ngerakines: Where did this dependency come from?

pages/login.tmpl
```diff
@@ -42,14 +46,17 @@ hr.short {
 		{{range .Flashes}}<li class="urgent">{{.}}</li>{{end}}
 	</ul>{{end}}
 
-	{{ if or .OauthSlack .OauthWriteAs }}
+	{{ if or .OauthSlack .OauthWriteAs .OauthGitlab }}
```
ngerakines: I'm not sure if the `or` template function accepts more than one parameter. It may have to be rewritten as:

```
{{ if or (or OauthSlack .OauthWriteAs) .OauthGitlab }}
```

oauth.go
```diff
@@ -165,6 +165,33 @@ func configureWriteAsOauth(parentHandler *Handler, r *mux.Router, app *App) {
...
+				callbackLocation: app.Config().App.Host + "/oauth/callback/gitlab",
+				httpClient:       config.DefaultHTTPClient(),
+			}
+			callbackLocation = app.Config().GitlabOauth.CallbackProxy
+		}
+
+		oauthClient := gitlabOauthClient{
+			ClientID:         app.Config().GitlabOauth.ClientID,
+			ClientSecret:     app.Config().GitlabOauth.ClientSecret,
+			ExchangeLocation: config.OrDefaultString(app.Config().GitlabOauth.TokenLocation, writeAsExchangeLocation),
```
ngerakines: The default value should be `gitlabExchangeLocation`.

oauth.go
```diff
@@ -165,6 +165,33 @@ func configureWriteAsOauth(parentHandler *Handler, r *mux.Router, app *App) {
...
+			InspectLocation:  config.OrDefaultString(app.Config().GitlabOauth.InspectLocation, writeAsIdentityLocation),
```
ngerakines: The default value should be `gitlabIdentityLocation`.

oauth.go
```diff
@@ -165,6 +165,33 @@ func configureWriteAsOauth(parentHandler *Handler, r *mux.Router, app *App) {
...
+			AuthLocation:     config.OrDefaultString(app.Config().GitlabOauth.AuthLocation, writeAsAuthLocation),
```
ngerakines: The default value should be `gitlabAuthLocation`.

go.mod
paddatrapper: No idea... I did nothing to change this file

pages/login.tmpl
paddatrapper: It seems to work as expected at the moment - if any one of the three are set, it shows the option

paddatrapper:
I have fixed the issues and can now successfully log in using GitLab OAuth as the authentication method

thebaer:
Some things I'm wondering about:

* Should we simplify configuration so you only need to set the Gitlab instance hostname? Then we could just add the path to the URL for each endpoint.
* Should we make the "Sign in with Gitlab" button text configurable? I'm thinking users might want to know which Gitlab instance they'd be authenticating with.

Thoughts?

paddatrapper:
> Should we simplify configuration so you only need to set the Gitlab instance hostname? Then the application could just append the path to the URL for each endpoint.

I would prefer this, but I wasn't sure if the API server could be a separate URL to the instance. Does anyone know if that is possible?

> Should we make the "Sign in with Gitlab" button text configurable? I'm thinking users might want to know which Gitlab instance they'd be authenticating with.

Good idea. Will do

thebaer:
> I would prefer this, but I wasn't sure if the API server could be a separate URL to the instance. Does anyone know if that is possible?

I'm not sure. But if we don't hear otherwise, I think we can go ahead and implement it that way now. If we pick a good name for the config field, we can always re-add the fields for specific endpoints in the future, if they turn out to be necessary.

> Good idea. Will do

Awesome, sounds good. Of course, it'd probably be good to fall back to some kind of default string if there's none configured.

ngerakines:
There isn't any reason it can't be separate. The only consideration is making sure that we're not hardcoding any gitlab.com URLs so people with private/self-hosted instances can use it.

paddatrapper:
> I'm not sure. But if we don't hear otherwise, I think we can go ahead and implement it that way now. If we pick a good name for the config field, we can always re-add the fields for specific endpoints in the future, if they turn out to be necessary.

Done

> Awesome, sounds good. Of course, it'd probably be good to fall back to some kind of default string if there's none configured.

Defaults to GitLab if not configured

config/config.go
```diff
@@ -69,6 +69,15 @@ type (
 		CallbackProxyAPI string `ini:"callback_proxy_api"`
 	}
 
+	GitlabOauthCfg struct {
+		ClientID         string `ini:"client_id"`
+		ClientSecret     string `ini:"client_secret"`
+		ServerAddress    string `ini:"server_address"`
```
thebaer: What do we think about using `host` as the field name, to echo other configuration fields?

config/config.go
```diff
@@ -69,6 +69,15 @@ type (
...
+		ServerName       string `ini:"server_name"`
```
thebaer: Maybe `InstanceName` / `instance_name` for this one? Or `DisplayName` / `display_name`? (I'd lean toward the latter.) _Server_ makes me think of the host / machine running Gitlab, and not of the software level or what the name is for.

paddatrapper:
Fixed

thebaer:
Thanks @paddatrapper and @ngerakines! Merging now.

PR #284, opened 2020-03-18
Show 503 page on blogs under high load

thebaer:
This acknowledges "too many connections" and "max user connections" errors in MySQL and propagates the error up the chain so we can notify the user and return the correct HTTP code.


thebaer:
Merging now.

Issue #286, opened 2020-03-21
Setup with reverse proxy leads to some static GET to localhost:[port] instead of the root domain


thepra:
### Describe the bug
As in the title, after setting up the writefreely server from the guide I found that out of the gate some resource were still pointing to localhost as if they were hard setup, like:
![image](https://user-images.githubusercontent.com/7032509/77233326-1688dc80-6ba7-11ea-88bf-215d42575803.png)

Bug number 2 is that error appearing in the h.js which you should look up(from Firefox latest release).

### Steps to reproduce (if necessary)
1. Setup writefreely with reverse proxy
2. Open the website
3. Look at the console for errors

### Expected behavior
What should've happened?
Right out of the box have pointing to static resources done right. And JS without error.

### Application configuration
 - **Single mode or Multi-user mode?** Multi-user
 - **Database?** sqlite
 - **Open registration?** no
 - **Federation enabled?** yes

**Version or last commit**: 0.11.2

thepra:
One solution could be that at the `writefreely --config` step there is also an option that asks if the writefreely server is going to be behind a reverse proxy, and if so, switch from pointers from localhost:[port] to blog.example.com

thebaer:
Thanks for bringing this up, @thepra. What is your `host` config value? If it's set to `localhost` and you intend this to be accessed outside the local machine, you'll want to instead set it to the public-facing URL ([see config docs](https://writefreely.org/docs/latest/admin/config#app) for more info). Based on your screenshots, it should probably be `https://blog.[redacted-domain]`

thebaer:
Closing for now. Please let us know if this turns out to be something other than a misconfiguration.

PR #287, opened 2020-03-21
Updated to address the issue with reverse proxy setup

thepra:
{{.Host}} should be fixed and become the domain chosen if the prod setup lives behind a proxy. The `Theme` is messed up because points to localhost:[port]
As of now it's not part of the installation process to setup the host in a way that supports reverse proxy handling/pointing.

---

### DIFF
--- a/templates/base.tmpl
+++ b/templates/base.tmpl
@@ -2,12 +2,12 @@
 <html>
 	<head>
 		{{ template "head" . }}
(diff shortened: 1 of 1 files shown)


thebaer:
As mentioned in [this comment](https://github.com/writeas/writefreely/issues/286#issuecomment-602101605), that issue was most likely caused by a configuration error instead of an application bug. Will close this for now, pending your response to my comment over on #286.

PR #288, opened 2020-03-27
Add CSS cache busting to templates in release

thebaer:
This adds a unique query string to the stylesheet's URL when doing a WriteFreely release, to ensure WF users get the latest styles after an admin upgrades their instance.


thebaer:
Merging now.

PR #289, opened 2020-03-28
Login with generic oauth feature

ketudb:
This aims to add a configurable, generic OAuth client module, so that it can be used with things like Keycloak etc, without needing a separate OAuth client configuration for every provider.

It adds the following ini settings under `oauth.generic`:
- `client_id`, `client_secret`, `CallbackProxy`, `CallbackProxyAPI` as usual
- `host`: The hostname/base URL for the oauth server
- `display_name`: The display name (defaults to "OAuth")
- `token_endpoint`, `inspect_endpoint`, `auth_endpoint` which will all be different depending on the oauth implementation.

---

### DIFF
--- a/account.go
+++ b/account.go
@@ -323,7 +323,7 @@ func viewLogin(app *App, w http.ResponseWriter, r *http.Request) error {
 		app.Config().WriteAsOauth.ClientID != "",
 		app.Config().GitlabOauth.ClientID != "",
         app.Config().GenericOauth.ClientID != "",
-        config.OrDefaultString(app.Config().GenericOauth.DisplayName, oAuthGenericDisplayName),
+        config.OrDefaultString(app.Config().GenericOauth.DisplayName, genericOauthDisplayName),
 		config.OrDefaultString(app.Config().GitlabOauth.DisplayName, gitlabDisplayName),
 	}
 
--- a/account.go
+++ b/account.go
@@ -66,7 +66,7 @@ func NewUserPage(app *App, r *http.Request, u *User, title string, flashes []str
 }
 
 func canUserInvite(cfg *config.Config, isAdmin bool) bool {
-	return cfg.App.UserInvites != "" &&
+      return cfg.App.UserInvites != "" &&
 		(isAdmin || cfg.App.UserInvites != "admin")
 }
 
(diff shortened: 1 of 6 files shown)


thebaer:
Thanks for submitting this, @ketudb! This looks really good so far, and will make a great addition to WriteFreely!

Just one thing that stands out so far -- we'll need to add _Connect_ / _Disconnect_ buttons to the Account Settings page. This functionality was added in #243 and originally left out of [my tutorial](https://discuss.write.as/t/feature-suggestion-login-with-an-account-from-other-websites-e-g-mastodon-github-etc/872/8), but I've just updated it. Once you get that part in, we'll do a full review.

ketudb:
> Just one thing that stands out so far -- we'll need to add _Connect_ / _Disconnect_ buttons to the Account Settings page.

Sure, I'll bounce back to that. How would you feel about that being optional, with the context that in some circumstances that it could be the _required_ way of logging in?

thebaer:
> How would you feel about that being optional, with the context that in some circumstances that it could be the _required_ way of logging in?

I'm absolutely open to that. Off the top of my head, I think it'd still be worth indicating that the account is connected in the UI, but just preventing people from disconnecting it. What do you think?

Also, I'm not sure if this'll require reworking normal username / password auth. If so, generally speaking, I'd just say that fewer configuration fields == better.

ketudb:
> Off the top of my head, I think it'd still be worth indicating that the account is connected in the UI, but just preventing people from disconnecting it. What do you think?

Hmm... yeah that makes sense. I'll have to dig into this again once I rebuild my dev instance.

> Also, I'm not sure if this'll require reworking normal username / password auth. If so, generally speaking, I'd just say that fewer configuration fields == better.

I do agree with you! I mostly see the use-case here is being SSO with Keycloak etc, which is the use-case I'm testing this change for/intend to use it for.

pascoual:
Thanks for this PR @ketudb!
I've tested it on a [Hiboo SSO](https://forge.tedomum.net/acides/hiboo) OIDC compliant and use scope "profile" but not "read_user". Can we imagine adding an option to specify the scope to ask ?
After login through SSO the default blog is missing: "This page is missing". Do you encounter the same ? If you want me to look deeper I can.

thebaer:
This is superseded by #317 -- closing this, and we'll finish up there.

PR #293, opened 2020-04-03
Added Gitea OAuth login and account management.

gytisrepecka:
Added [Gitea OAuth login and account management](https://discuss.write.as/t/gitea-oauth/1235) functionality based on [Matt's guidelines](https://discuss.write.as/t/feature-suggestion-login-with-an-account-from-other-websites-e-g-mastodon-github-etc/872/8).

---


oauth_gitea.go
```diff
@@ -0,0 +1,115 @@
...
+	ExchangeLocation string
+	InspectLocation  string
+	CallbackLocation string
+	HttpClient       HttpClient
+}
+
+var _ oauthClient = giteaOauthClient{}
+
+const (
+	giteaHost        = "https://source.gyt.is"
```
cjeller1592, line 24: Could you add `giteaHost` into the `giteaOauthClient` struct along with a function to retrieve it for later use in the file? Maybe call it "host"? That way all the set up needed for Gitea OAuth is in the `config.ini` file rather than across multiple files — just add the necessary values to the config file and you're done!

cjeller1592:
Got it working on my instance @gytisrepecka! Just had an issue during set up that I lay out in a [comment](https://github.com/writeas/writefreely/pull/293/commits/c798a44f697c69f394113e4fb8c5f2bfc268c839#r409752077). Otherwise works great from a user perspective!

oauth_gitea.go
gytisrepecka: My bad, forgot the hardcoded value... Thanks, will submit a fix!

gytisrepecka: Fixed with following steps:

1. From `oauth_gitea.go` removed `giteaHost` variable - no need to have it as Gitea doesn't have hosted service (like GitLab does):

https://github.com/writeas/writefreely/pull/293/commits/658310bc24d6da87e7e27a8c5d870399ee4277e8

2. In `oauth.go` modified function `configureGiteaOauth`: no need to use additional variable `address` because to form `Location`s URLs we can use value from `config.ini` right away - it is already available in `app.Config().GiteaOauth.Host` (`host` from `oauth.gitea` section in `config.ini` file):

https://github.com/writeas/writefreely/pull/293/commits/0e1459c6b251f95b376c548d37a78dce5c3571ef

@cjeller1592 , do you think it is worth adding `Host` to `giteaOauthClient` struct? It anyway has full URLs in `Location`s, so maybe it doesn't make sense to store hostname separately?

https://github.com/gytisrepecka/writefreely/blob/0e1459c6b251f95b376c548d37a78dce5c3571ef/oauth_gitea.go#L11

cjeller1592: As long as adding the necessary values to `config.ini` is all that has to be done, then remove (or leave) as needed. We just want setting up Gitea OAuth to be as easy as possible without having to go into other parts of the codebase.

gytisrepecka: Certainly no configuration should reside in source.

If you have a chance, pull oauth-gitea and try out again - I've checked if config.ini values are considered after removing that hardcoded hostname.

One more thing I've noticed: removal of OAuth provider from account uses not use display_name but generic name of that login type, so would show Gitea regardless of display_name set - same story with other implementations. But signing up and adding OAuth to existing account uses display_name from confog.ini. Will look into it :)

gytisrepecka:
If possible, pull sources once again and try building:
```
go get -d -u -v code.gyt.is/writefreely
cd ~/go/src/code.gyt.is/writefreely
git checkout oauth-gitea
```

cjeller1592:
> If possible, pull sources once again and try building:
> 
> ```
> go get -d -u -v code.gyt.is/writefreely
> cd ~/go/src/code.gyt.is/writefreely
> git checkout oauth-gitea
> ```

Just did this & got Gitea OAuth login working on my instance with the added `host` in the config! Thanks for the additions @gytisrepecka!

gytisrepecka:
Merged current `develop` branch and resolved `pages/login.tmpl` conflict.

This version is running live on [fedi.dev](https://fedi.dev/).

thebaer:
Thanks for your patience on my review -- just some quick formatting needed and we can get this in.

gytisrepecka:
Performed formatting:
```bash
$ gofmt -w account.go
$ gofmt -w config/config.go
$ gofmt -w oauth.go
$ gofmt -w oauth_gitea.go
$ gofmt -w routes.go
```
Please take a look if there's something else.

PR #295, opened 2020-04-09
Optimize Drafts retrieval

thebaer:
This adds a database index to speed up retrieval of Drafts.

Most small instances won't notice any difference from this change, but it'll ensure they stay fast as they grow. It was driven by the performance issues we've encountered on Write.as as we've approached 800,000 posts on the platform. Response times for our "Drafts" page (labelled as "Posts") dropped from several seconds to under half a second with this change.

~~This is a draft, since it's untested with SQLite.~~ Ready for review now.


thebaer:
Merging now.

Issue #297, opened 2020-04-12
doesn't support encrypted database connections


fcevado:
### Describe the bug
if you start MySQL with `--ssl` mode enabled writefreely won't connect to the database.

### Steps to reproduce (if necessary)
1. start MySQL with `--ssl` mode.
2. run `writefreely --config` and set the proper database credentials
3. writefreely will error try to connect to the databse with `this authentication plugin is not supported`

### Expected behavior
What should've happened?
During the config I'd expect it to ask if it will require ssl mode or not and be able to work properly with it.

### Application configuration
 - **Single mode or Multi-user mode?** Multi-user mode 
 - **Database?** [mysql/sqlite] MySQL
 - **Open registration?** [yes/no] no
 - **Federation enabled?** [yes/no] yes

**Version or last commit**:
0.11.2

thebaer:
Thanks for reporting this! A simple fix will probably involve:

* Adding a new [`DatabaseCfg`](https://github.com/writeas/writefreely/blob/e51e58386ebd7299f5f18de1ab6aee6c057f968c/config/config.go#L52-L60) flag that enables TLS / SSL
* If that flag is true, include [`&tls=true`](https://github.com/go-sql-driver/mysql#tls) in the database connection string: https://github.com/writeas/writefreely/blob/e51e58386ebd7299f5f18de1ab6aee6c057f968c/app.go#L751
  * (Really, you can include `&tls=true` or `&tls=false`, where true / false is the value of the config flag)

This is a pretty straightforward fix, so please feel free to submit a pull request for it!

PR #301, opened 2020-04-16
fix: #297 - tls for mysql connections

joicemjoseph:
#297 - tls for mysql connections.
---


thebaer:
Thanks for submitting this! Looks good after reading through -- just need to test and confirm.

PR #303, opened 2020-04-20
Respect registration config on OAuth signup flow

thebaer:
This alters the current OAuth flow to respect an instance's user registration config. It includes:

* Closed registration. Now, if the OAuth flow would've normally led to the creation of a new user on a closed-registration instance, it is blocked instead and an error is returned to the user.
* Invite-based registration. Now, on an invite link, a user has the option of signing up via any configured OAuth provider.

This PR includes style and database changes. Run the following commands to update your development environment:

```
make ui
writefreely db migrate
```

### DIFF
--- a/account.go
+++ b/account.go
@@ -489,6 +489,9 @@ func login(app *App, w http.ResponseWriter, r *http.Request) error {
 				return impart.HTTPError{http.StatusPreconditionFailed, "This user never added a password or email address. Please contact us for help."}
 			}
 		}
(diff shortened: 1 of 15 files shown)


thebaer:
Added a few light improvements to the OAuth signup flow here, too:

* Skip random username generation if unnecessary
* Use "sign up" language on OAuth signup page, instead of "log in" language
* Inform user when they try to login to an OAuth-based account with a password but don't have one set

PR #307, opened 2020-04-28
Support hashtags with numbers + letters

thebaer:
Previously, a hashtag like [#100DaysToOffload](https://100daystooffload.com/) wouldn't automatically turn into a clickable link -- all hashtags needed to _start_ with an alphabetic character.

The [saturday](https://github.com/writeas/saturday) Markdown library we use now also parses out hashtags that begin with a number AND contain letters (but not numbers only). So this PR updates that dependency to bring the fix into WriteFreely.


thebaer:
Confirmed to be working well. Merging now.

Issue #217, opened 2019-11-30
Header 3 (###) is same size as Header 2 (##)


cjeller1592:
### Describe the bug
When you use the Header 3 tag (###), it is the same size as the Header 2 tag (##) on both single post and collection view.

Example Post: https://write.as/2dh764zy7nga29yw.md

Topic on Forum where issue was brought up: https://discuss.write.as/t/any-reason-why-headings-two-and-three-are-the-same-size/959

### Expected behavior
Header 3 should be smaller than Header 2, not the same size.

PR #315, opened 2020-05-25
core.less: Use a bigger font size for h2

RJ722:
The font-size for h2 was set to be 1.17em, which is also the font size defined for h3 by user agent stylesheet. This lead to both h2 and h3 being rendered in same size. Use 1.2em for h2.

Closes: #217

---

### DIFF
--- a/less/core.less
+++ b/less/core.less
@@ -81,7 +81,7 @@ body {
 				font-size: 1.5em;
 			}
 			h2 {
-				font-size: 1.17em;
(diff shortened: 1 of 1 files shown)


thebaer:
Thanks for contributing this, @RJ722! I'll test and verify this fix soon. Otherwise, anyone else should feel free to weigh in!

thebaer:
Making these revisions in #368.

PR #317, opened 2020-06-06
Login with generic oauth feature++

pascoual:
**Create from @ketudb [PR](https://github.com/writeas/writefreely/pull/289) with @thebaer requested modifications about Connect / Disconnect buttons from Account Settings page**  

This aims to add a configurable, generic OAuth client module, so that it can be used with things like Keycloak etc, without needing a separate OAuth client configuration for every provider.

It adds the following ini settings under `oauth.generic`:

    * `client_id`, `client_secret`, `CallbackProxy`, `CallbackProxyAPI` as usual

    * `host`: The hostname/base URL for the oauth server

    * `display_name`: The display name (defaults to "OAuth")

    * `token_endpoint`, `inspect_endpoint`, `auth_endpoint` which will all be different depending on the oauth implementation.
   
    * `allow_disconnect`: allow the user to disconnect his writefreely account from `oauth.generic` one

I've try to contact @ketudb but without success.


thebaer:
Thanks for taking over development on this, @pascoual! I'll review this soon.

ketudb:
Apologies for dropping out on this one - got extremely busy elsewhere. @pascoual thank you for jumping in here, and I'm sorry you were unable to reach me. I'll try to keep an eye on this PR - if you need anything mention me and I'll try to jump in :smile:

pascoual:
I need to add an option : disable_password_auth as we want to use WF with the SSO only.
@thebaer: I will make few more commits

ketudb:
> I need to add an option : disable_password_auth as we want to use WF with the SSO only.
> @thebaer: I will make few more commits

This would be awesome, since it's _exactly_ what I was trying to do. Thanks :grin:

pascoual:
Look at the last commit, it's done ;) !

ketudb:
@pascoual that’s super awesome! thank you for taking over from me - this is not my coding language haha

i can’t wait to try it out

c-cesar:
This would enable to use WriteFreely on multiuser instances of YuNoHost, using one other app as oauth provider, like Nextcloud, without the need to implement LDAP.
And this is my current usercase. 
Tank you so much.

pages/landing.tmpl
```diff
@@ -73,6 +78,22 @@ form dd {
 
 	<div{{if not .OpenRegistration}} style="padding: 2em 0;"{{end}}>
 		{{ if .OpenRegistration }}
+		{{if .DisablePasswordAuth}}
```
thebaer, line 81: With this logic, admins won't be able to allow **both** password and OAuth account creation. So I think we'll want to repeat the logic from `login.tmpl` here, and simply hide the password-based signup form when `DisablePasswordAuth == true`, instead of making the signup methods mutually exclusive.

pascoual:
Got it, will fix it asap.

pascoual:
I've also fix some merging issues from previous commit and find running WF in dev.

PR #319, opened 2020-06-08
Support AP-style mentions of centralized social media accounts

thebaer:
This allows users to easily link to users on the following non-ActivityPub platforms:

- deviantart.com
- facebook.com
- flickr.com
- github.com
- instagram.com
- medium.com
- reddit.com
- twitter.com
- wattpad.com
- youtube.com

Of course, since these platforms don't speak ActivityPub, mentioned users won't be notified. But for example, adding `@ilikebeans@twitter.com` in a post will send users to the correct Twitter profile when clicked.

This also adds missing error handling in `federatePost()`.

What does everyone think about adding this? Would love to hear your feedback.


shinenelson:
I'm not sure whether this would categorize as a whole separate feature in itself, but I'm going to put it out there anyway.

How about implementing IndieWeb [POSSE](https://indieweb.org/POSSE) and [backfeed](https://indieweb.org/backfeed)?

I'm aware that this is way beyond the scope of what this pull request is trying to do. I'm probably coming out of nowhere and stepping on your toes, but I thought it'd be a good idea that could also pave the way for a bigger discussion.

And to start the discussion, I'd say the advantage of implementing POSSE would be that the syndication of posts would be uniform everywhere and they'll all point back to one canonical URL.

And _IF_, at some point in the future, [Twitter](https://www.theverge.com/2019/12/12/21012553/twitter-bluesky-decentralized-social-network-developers-reaction-mastodon-activitypub) or other social media decide to embrace ActivityPub, this would still work backward compatible too.

rgx1:
I think this is a great idea. 

I do use a couple of different Crossposters, and one of them, [the one made by Renato Lond](https://crossposter.masto.donte.com.br/), does a beautiful job of bridging the platforms.

I have mentioned [them both in a recent post](https://write.tedomum.net/rgx/crossposting-and-surprises-happened). 

------
Renato's output will look like this :  

image

The link to the source post is live and working. Another crossposter I use does less well, creating less attractive output.

image

...although the link to source is present and working.

-------

We could do better yet -- I could created a better looking post, with the full source URL hidden (and it will display so, even in the normal Mastodon instances), using the MarkDown support provided at my Glitch.Social instance. 

Source looks like :

```
#xp

RT [@TheMedicaIVideo](https://twitter.com/TheMedicaIVideo/status/1266604854358814722)

the beauty of brown eyes 👁`
```

And outputs : 

image

[That post is Here.](https://muensterland.social/@rgx/104310624318785578) 

------

As I noted earlier this week upon experimentation, this looks wonderful on Glitch-Social, and will display hidden URL even on normal Mastodon 3.x.x instances.

There seems to be an allowance for these hidden URLs in posts originating from other platforms in Fediverse -- and while plain vanilla Mastodon **WILL strip off Bold, Italics,** etc, it *will preserve the URLs as intended*.

I love this -- and have made posts now that have up to 5 URLs included and not a single 'https' to be seen. Clean, elegant, functional.

thebaer:
Thanks for the ideas! This is definitely a larger discussion that's out of scope for this PR, so let's move it to [this thread](https://discuss.write.as/t/posse-and-backfeed/1510) on our forum -- the right place for these kinds of discussions.

shinenelson:
agreed. I wasn't aware of the forum; or I didn't look enough. I should've just assumed that it existed though and looked for it because every community has a discourse these days :smile:

thebaer:
Moved this functionality to the `web-core/silobridge` package, and added support for more sites:

- deviantart.com
- facebook.com
- flickr.com
- github.com
- instagram.com
- reddit.com
- wattpad.com
- youtube.com

Merging now!

PR #324, opened 2020-06-11
Show warning in editor when local draft is out of date

### LINKED ISSUE #41: After update through API, editor shows old content
When updating a post through the API while also checking the editor on the website, the editor does not use the updated content which came in, but seems to be using a local browser storage entry `draft......` to show the editable content.

Steps to reproduce the behavior (I have used the writefreely.el package):
1. Create a post through the api as draft
2. Got to the site, find the post and edit
3. Do other things on site
4. Post update through api
5. Go back to post and edit

Editor will show initial content created in 1. or, if applicable any changes made in the editor in step 2. (but not published)

### Expected behavior
Post should be updated according to latest update through api. It looks as if the local storage does not get invalidated properly.

### Application configuration
- Multi-user / open reg / federation enabled
- latest git version

thebaer:
When editing a post, this tracks the `updated` date that the local, in-browser draft is based on, and compares it to the post's actual `updated` value in the database.

If the database's `updated` value is after the local draft's `updated` date, this warning is shown to the user, allowing them to delete their local draft and load the latest version:

Screenshot from 2020-06-11 11-46-46

This fixes #41.

**Testing notes**: This PR includes style changes; update your stylesheet by running: `make ui`. Also note your database timezone **must** be set to UTC for this to work correctly.

---

### DIFF
--- a/less/pad.less
+++ b/less/pad.less
@@ -361,6 +361,24 @@ body#pad {
     z-index: 10;
 }
 
+body#pad .alert {
+	position: fixed;
+	bottom: 0.25em;
+	left: 2em;
+	right: 2em;
+	font-size: 1.1em;
+
+	&#edited-elsewhere {
+		&.hidden {
+			display: none;
+		}
+
+		a {
+			font-weight: bold;
+		}
+	}
+}
+
 @media all and (max-height: 500px) {
 	body#pad {
 		textarea {
@@ -425,6 +443,10 @@ body#pad {
 			padding-left: 10%;
 			padding-right: 10%;
 		}
+		.alert {
+			left: 10%;
+			right: 10%;
+		}
 	}
 }
 @media all and (min-width: 60em) {
@@ -432,6 +454,10 @@ body#pad {
 		textarea {
 			padding-left: 15%;
 			padding-right: 15%;
+		}
+		.alert {
+			left: 15%;
+			right: 15%;
 		}
 	}
 }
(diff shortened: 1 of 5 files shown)


cjeller1592:
Looks good on my end! 

Would it be worth it to add language mentioning _if_ you want to override the draft updated elsewhere?

>This post has been updated elsewhere since you last published! If you want to override that update, then continue with this draft. If not, **delete draft and reload**.

But that might be such a _small_ use case that I don't think it's worth mentioning.

thebaer:
Thanks for checking this out, @cjeller1592! I wanted to keep it concise, which is why I chose the current verbiage. But if it's confusing to people we can definitely change it in the future.

Merging this now!

sobodash:
I recently updated to the new release candidate and am observing an odd behavior with this feature. All posts show this message, and clicking delete and redraft simply creates a new blank post that still shows this message.

I would guess this must be related to time configuration. I checked and the machine is set to UTC and is synced with a time server. My local machine is configured to the same time in EST.

Any ideas what might be going on?

thebaer:
Hmm, what timezone is your database set to (assuming you're running MySQL)?

sobodash:
> Hmm, what timezone is your database set to (assuming you're running MySQL)?

I don't see anything setting a timezone in /etc/mysql. I suppose it would be whatever the default is after running the MariaDB setup scripts on a stock Ubuntu machine set to UTC.

Since I never explicitly set up time zones in SQL before, I just ran `mysql_tzinfo_to_sql /usr/share/zoneinfo | sudo mysql -u root -p mysql` and got the output:

> Warning: Unable to load '/usr/share/zoneinfo/leap-seconds.list' as time zone. Skipping it.
> Warning: Unable to load '/usr/share/zoneinfo/leapseconds' as time zone. Skipping it.
> Warning: Unable to load '/usr/share/zoneinfo/tzdata.zi' as time zone. Skipping it.

PR #327, opened 2020-06-21
Add CORS header ('Access-Control-Allow-Origin') to public APIs

Dar13:
This permits external websites to query WriteFreely instances for:
 * Collections
 * Public posts in collections
 * Specific public posts.

Without these changes, websites attempting to use the public API to query collections (such as my site: [moor3.xyz](moor3.xyz), [source](https://gitlab.com/Dar13/moor3-website)) will receive CORS errors. @cjeller1592 helped me in the past with a CORS proxy ([discussion](https://discuss.write.as/t/javascript-cors-errors/810?u=dar13)), but it's time to fix it at the source.

I consider this PR to be more like an RFC, I'm fully open to guidance/critique on how to make these changes acceptable.

---

### DIFF
--- a/handle.go
+++ b/handle.go
@@ -599,6 +599,9 @@ func (h *Handler) AllReader(f handlerFunc) http.HandlerFunc {
 				log.Info(h.app.ReqLog(r, status, time.Since(start)))
 			}()
 
+			// Allow any origin, as public endpoints are handled in here
+			w.Header().Set("Access-Control-Allow-Origin", "*");
+
 			if h.app.App().cfg.App.Private {
 				// This instance is private, so ensure it's being accessed by a valid user
 				// Check if authenticated with an access token


thebaer:
Thank you for taking this on! I agree we should fix this in the application, and these are probably the most important API endpoints to do this on. 

One thing that jumps out at me: as it stands, the header will only be set on successful responses. If an `error` is returned in any of these funcs, it bubbles up to the `Handler.All()` middleware and the `Access-Control-Allow-Origin` header is never set.

So a better place for these additions might be in the `Handler.All()` method (or a modified version, or a new variation of it), somewhere before this line: https://github.com/writeas/writefreely/blob/5c94d23466dc3e7e1821b845d9d7db9294c48af4/handle.go#L541

Dar13:
> So a better place for these additions might be in the Handler.All() method (or a modified version, or a new variation of it), somewhere before this line

It seems `AllReader` is a suitable place as it's only `GET` requests and the endpoints using it are (as best as I can tell) endpoints that don't require authentication to access if the instance is public. I've adjusted the patch in this PR accordingly.

For authenticated endpoints I think a more fully-fleshed out approach is needed, such as users choosing which origin(s) are permitted to query/update/post/etc among others.

thebaer:
Sorry for the delay on this! I agree, the `AllReader` method is a good place for the header that should cover most basic needs.

Beyond that, we can definitely expand on this with better customization in the future. If you'd like to get that discussion started, please feel free to start a topic [on the forum](https://discuss.write.as/c/development/7)!

Merging now -- thanks again for contributing!

PR #330, opened 2020-06-23
Support post signatures

thebaer:
This enables users to add a signature to all blog posts, and update it from a single location (the blog's "Customize" page).

This change requires a database migration. Run with:

```
writefreely db migrate
```

Closes [T582](http://writefreely.org/tasks/582)

---


cjeller1592:
Looks good on my end! I was able to add post signatures and have them show up on my blog. Only thing I'd add is that the "shortcodes" part in the signature description might confuse people. Of course that's at this present moment — when shortcodes are added for things other than _Read more..._ it will make sense.

thebaer:
Thanks, @cjeller1592! Yep, we'll add and link to needed documentation when that becomes an issue.

Merging now!

PR #333, opened 2020-07-02
Create dependabot.yml

shleeable:
Github Dependabot will create PRs for go modules on a weekly basis.

Example from my fork:
image

---

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -2,5 +2,6 @@ version: 2
 updates:
   - package-ecosystem: "gomod" # See documentation for possible values
     directory: "/" # Location of package manifests
+    open-pull-requests-limit: 50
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,7 @@
+version: 2
+updates:
+  - package-ecosystem: "gomod" # See documentation for possible values
+    directory: "/" # Location of package manifests
+    open-pull-requests-limit: 50
+    schedule:
+      interval: "weekly"
```
thebaer, line 7: I think `monthly` would be more manageable for us, since this is going to require regular testing.

shleeable:
> Thanks for opening this! I think this would be a nice addition -- just one change I'd suggest.

done

PR #336, opened 2020-07-21
Add Cache-Control header

### LINKED ISSUE #322: static assets served without cache policy leads to suboptimal page speed
### Describe the bug
Static assets (list below) are served without any caching header, which results in the browser re-requesting the files too often.

> /fonts/Lora-Bold.woff2
> /fonts/Lora-Regular.woff2
> /fonts/open-sans-v13-latin-regular.woff2
> /js/webfont.js
> /js/h.js
> /js/postactions.js
> /js/localdate.js
> /css/fonts.css

1. Open Browser (e.g., Firefox) Developer Tools and its network panel
2. Load a writefreely instance
3. Observe response headers

### Expected behavior
writefreely server should have responded with a cache header

### Application configuration
 - **Single mode or Multi-user mode?** single
 - **Database?** [mysql/sqlite] mysql

**Version or last commit**:
v0.12.0

Antolius:
Add Cache-Control response header for static files in order to improve page loading speed.
Fixes  #322

---

### DIFF
--- a/cache.go
+++ b/cache.go
@@ -11,6 +11,7 @@
 package writefreely
 
 import (
+	"net/http"
 	"sync"
 	"time"
 )
@@ -66,4 +67,11 @@ func GetPostsCache(userID int64) *[]PublicPost {
 		return nil
 	}
 	return pci.Posts
+}
+
+func cacheControl(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Cache-Control", "public, max-age=604800, immutable")
+		next.ServeHTTP(w, r)
+	})
 }
--- a/cache.go
+++ b/cache.go
@@ -11,7 +11,6 @@
 package writefreely
 
 import (
-	"net/http"
 	"sync"
 	"time"
 )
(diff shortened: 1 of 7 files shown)


Antolius:
Hi! I'm new to the code-base, so might have missed something obvious, but this seems rather straightforward.

I've also noticed a TODO comment on the [`InitStaticRoutes` method](https://github.com/writeas/writefreely/blob/develop/routes.go#L25-L26) that says it should be converted into a func. If you'd like I can just update the method into something like this:
```go
func InitStaticRoutes(app *App, r *mux.Router)
 ```
and leave its implementation as-is.

Or, if the comment had some  broader implications I could try working on them as well. For example, it seems like [`shttp` field in the App struct](https://github.com/writeas/writefreely/blob/develop/app.go#L68) could be removed, in which case the `InitStaticRoutes` function would not need to modify the app passed into it. This might be a bit bigger change, so it might be better to have it in a separate pull request.

File: cache.go [hunk 1]
```diff
@@ -67,3 +68,10 @@ func GetPostsCache(userID int64) *[]PublicPost {
 	}
 	return pci.Posts
 }
+
+func cacheControl(next http.Handler) http.Handler {
```
thebaer, line 72: Since `cache.go` is meant more for internal application caches, I'd suggest moving this to a place where we have other HTTP middleware, like `handle.go`.

Antolius: Sure thing! I moved it in [this commit](https://github.com/writeas/writefreely/pull/336/commits/1b1d3064c9064ca76ee629d07ff79fbdaffd31ca#diff-fe997d9c2c36e76b19925e7f5fad6e48).

thebaer (approved):
Awesome, thanks!

PR #360, opened 2020-08-04
Add user silencing to CLI

mrvdb:
Add user silencing to CLI

Quick implementation. Ideally unsilence should be there as well and for scriptability a 'no-confirm' flag of some sort would be nice.

### DIFF
--- a/app.go
+++ b/app.go
@@ -747,6 +747,51 @@ func DoDeleteAccount(apper Apper, username string) error {
 	return nil
 }
 
+func DoSilenceAccount(apper Apper, username string) error {
+	// Connect to the database
+	apper.LoadConfig()
+	connectToDatabase(apper.App())
+	defer shutdown(apper.App())
+
+	// check user exists
+	u, err := apper.App().db.GetUserForAuth(username)
+	if err != nil {
+		log.Error("%s", err)
+		os.Exit(1)
+	}
+	userID := u.ID
+
+	// do not silence the admin account
+	// TODO: check for other admins and skip?
+	if u.IsAdmin() {
+		log.Error("Can not silence admin account")
+		os.Exit(1)
+	}
+
+	// confirm deletion, w/ w/out posts
+	prompt := promptui.Prompt{
+		Templates: &promptui.PromptTemplates{
+			Success: "{{ . | bold | faint }}: ",
+		},
+		Label:     fmt.Sprintf("Really silence user : %s", username),
+		IsConfirm: true,
+	}
+	_, err = prompt.Run()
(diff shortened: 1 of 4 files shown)


cmd/writefreely/main.go
```diff
@@ -85,6 +85,11 @@ func main() {
 				Usage:  "Delete a user with the given username",
 				Hidden: true,
 			},
+			&cli.StringFlag{
+				Name:   "silence-user",
```
thebaer, line 89: We kept old flags in here only for backwards compatibility. Ideally we'd move away from this older style, so we don't need to add new commands here.

cmd/writefreely/main.go
```diff
@@ -152,6 +157,8 @@ func legacyActions(c *cli.Context) error {
 		return writefreely.CreateUser(app, username, password, false)
 	case c.IsSet("delete-user"):
 		return writefreely.DoDeleteAccount(app, c.String("delete-user"))
+	case c.IsSet("silence-user"):
```
thebaer, line 160: Same here -- no need to check for this flag.

app.go
```diff
@@ -747,6 +747,51 @@ func DoDeleteAccount(apper Apper, username string) error {
...
+		IsConfirm: true,
+	}
+	_, err = prompt.Run()
+	if err != nil {
+		log.Info("Aborted...")
+		os.Exit(0)
+	}
+
+	log.Info("Silencing...")
+	err = apper.App().db.SilenceAccount(userID)
```
thebaer, line 786: I think ideally we should use the existing `datastore.SetUserStatus()` method instead of creating a new one here, to keep things maintainable.

mrvdb:
Thanks for the review Matt. Agreed on all points. It's hard to deduce the direction of code from existing stuff, so I took the lazy route and just did a 'copy-pasta' first to get it to work and allow me to manage some abuse on qua.name.

I'll be returning to this PR after returning from travel. I'll set state to WIP after I find out where to click.

thebaer:
No problem, @mrvdb. Any progress on this?

mrvdb:
Haven't been  revisiting it yet.

thebaer:
@mrvdb I'm hoping to get the next release out within the next week or two, so if you can make those changes before then, we'll include them in v0.13! Otherwise we can include it in the following release.

mrvdb:
Sorry, I  won't have the opportunity in the foreseeable future to spent time on this. Feel free to remove or complete as you see fit.

thebaer:
No problem, thanks for letting me know!

PR #364, opened 2020-08-07
Fix removal of query parameters on youtube embed links

### LINKED ISSUE #328: Embedded YouTube videos cannot start at a certain time using "?start=..."
### Describe the bug
This code https://github.com/writeas/writefreely/blob/develop/postrender.go#L100-L102 (which is states the intention is to remove `autoplay`) makes it so that I cannot have YouTube videos start at a precise moment using `?start=...`.

### Expected behavior
It should just remove the `autoplay` query parameter if that is the intention.  Although I don't understand why that would happen on the blog page.  It seems like something you would only want to remove on blog previews when there could be many annoying "autoplay"s triggered.

### Application configuration
 - **Multi-user mode** 
 - **Database?** sqlite
 - **Open registration?** yes
 - **Federation enabled?** no

**Version or last commit**: 457051106d9db8e739f7bfb8db878aeda9fb8e80

Obayanju:
This makes sure only the autoplay parameter in any youtube URL is disabled.

Fixes #328 

---


postrender.go
```diff
@@ -73,6 +74,25 @@ func applyMarkdown(data []byte, baseURL string, cfg *config.Config) string {
 	return applyMarkdownSpecial(data, false, baseURL, cfg)
 }
 
+func disableYoutubeAutoplay(outHTML string) string {
+	for _, match := range youtubeReg.FindAllString(outHTML, -1) {
+		u, err := url.Parse(match)
+		if err != nil {
+			log.Error("Couldn't parse youtube url: %v", err)
```
thebaer, line 81: If there's an error, we should probably move on to the next match (by adding a `continue` statement here).

postrender.go
```diff
@@ -73,6 +74,25 @@ func applyMarkdown(data []byte, baseURL string, cfg *config.Config) string {
...
+		}
+		u.RawQuery = html.UnescapeString(u.RawQuery)
+		q := u.Query()
+		// Set Youtube autoplay url parameter, if any, to 0
+		if len(q["autoplay"]) == 1 {
+			q.Set("autoplay", "0")
+		}
+		u.RawQuery = q.Encode()
+		clean_url := u.String()
```
thebaer, line 90: Style-wise, we use _camelCase_ instead of _snake_case_, so I'd recommend renaming this to e.g. `cleanURL`.

postrender.go
Obayanju: Thanks for the comment! I Will update that.

thebaer (approved):
Looks good, thanks for fixing those! Merging now.

PR #368, opened 2020-08-13
Use a bigger font size for h2

thebaer:
Continues work started in #315. From that issue:

The font-size for h2 was set to be 1.17em, which is also the font size defined for h3 by user agent stylesheet. This lead to both h2 and h3 being rendered in same size. Use ~~1.2em~~ 1.4em for h2.

Closes: #217

---


thebaer:
Merging now.

PR #369, opened 2020-08-19
Support Web Monetization

thebaer:
This PR adds support for [Web Monetization](https://webmonetization.org/).

This feature is optional -- admins can enable it by setting `monetization = true` in the `[app]` section of their config -- or changing the setting in their admin panel.

When enabled, users will find a new "Web Monetization" field on the Customize page, with a little information on how it works.

This closes [T773](https://phabricator.write.as/T773).

- [x] Support `monetization` meta tag
- [x] Enable end users to add a monetization pointer
- [x] Enable end users to update their monetization pointer

---

### DIFF
--- a/account.go
+++ b/account.go
@@ -839,6 +839,9 @@ func viewEditCollection(app *App, u *User, w http.ResponseWriter, r *http.Reques
 		return ErrCollectionNotFound
 	}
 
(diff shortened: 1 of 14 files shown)


thebaer:
Merging now!

PR #371, opened 2020-08-19
Navigation improvements

thebaer:
This PR brings some general quality-of-life improvements to blog navigation for multi- and single-user instances.

**Persistent "New Post" button in the global navigation bar**. Now it's easy to navigate back to the editor in a consistent way. Also, when on blog sub-sections like _Customize_ and _Stats_, the button is contextual, i.e. clicking it will automatically select the current blog as the publishing target in the editor. Previously, this button was only enabled when an instance was configured with `chorus = true`.

**Standard navigation across blog sub-sections**. Now you can easily switch between customizing your blog, viewing stats, and viewing the blog itself with standard navigation. On single-user instances, these sections have been moved to the global navigation bar.

### Customize blog page

| Before | After |
| --- | --- |
| Screenshot from 2020-08-19 18-14-05 | Screenshot from 2020-08-19 18-12-26|

### Customize blog page

| Before | After |
| --- | --- |
| Screenshot from 2020-08-19 18-17-05 | Screenshot from 2020-08-19 18-16-24 |

### Blogs page

| Before | After |
| --- | --- |
| Screenshot from 2020-08-19 18-15-04 | Screenshot from 2020-08-19 18-15-59 |

---

### DIFF
--- a/account.go
+++ b/account.go
@@ -49,6 +49,7 @@ type (
 		Separator template.HTML
 		IsAdmin   bool
 		CanInvite bool
+		CollAlias string
 	}
 )
 
@@ -854,6 +855,7 @@ func viewEditCollection(app *App, u *User, w http.ResponseWriter, r *http.Reques
 		Collection: c,
 		Silenced:   silenced,
 	}
+	obj.UserPage.CollAlias = c.Alias
 
 	showUserPage(w, "collection", obj)
 	return nil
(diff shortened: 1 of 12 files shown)


rkingett:
Make sure the <nav> HTML elements are used to denote landmarks and otherwise. But this all looks good! I'm hoping accessibility is retained. I'm also hoping [These HTML structural elements](http://web.simmons.edu/~grovesd/comm244/notes/week3/structural-elements) can be added, as well.

thebaer:
Thanks for your input! This does put blog navigation into the more semantic `nav` element (instead of `h4`). For any structural issues, I think we'll address them as they come up while making other accessibility improvements.

Merging now!

rkingett:
There is a little bit of excessive verbiage after navigating past the first banner landmark. The screen reader announces "Navigation landmark" 3 times.

Perhaps wrapping all the top elements into one HTML NAV would be good.

PR #382, opened 2020-09-05
Fix escaping in post summary

### LINKED ISSUE #340: Meta tags are double html-encoded
### Describe the bug

If you check the HTML source [on this post](https://mikestone.me/twitter-got-hacked-is-mastodon-immune), the HTML <meta> tags are double HTML encoded. For example:

```html
<meta property="og:description" content="Unless you&amp;#39;ve been hiding under a rock the last week or so, you probably know that Twitter got &amp;#34;hacked&amp;#34;. So, is Mastodon immu..." />
```

### Expected behavior

The `&amp;#34;` should be `&#34;`

### Application configuration
 - **Single mode or Multi-user mode?** multi-user
 - **Database?** n/a
 - **Open registration?** n/a
 - **Federation enabled?** n/a

**Version or last commit**: v0.12

Antolius:
Unescaping post content after sanitizing it. This will prevent double
escaping when summary is rendered by html/template package which does
escaping by default.
Fixes #340

---

### DIFF
--- a/postrender.go
+++ b/postrender.go
@@ -140,9 +140,7 @@ func applyBasicMarkdown(data []byte) string {
 func postTitle(content, friendlyId string) string {
 	const maxTitleLen = 80
 
-	// Strip HTML tags with bluemonday's StrictPolicy, then unescape the HTML
-	// entities added in by sanitizing the content.
-	content = html.UnescapeString(bluemonday.StrictPolicy().Sanitize(content))
+	content = stripHTMLWithoutEscaping(content)
 
 	content = strings.TrimLeftFunc(stripmd.Strip(content), unicode.IsSpace)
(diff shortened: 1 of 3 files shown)


Antolius:
The double encoding issue could also be fixed by updating templates to use the `SummaryHTML` method [here](https://github.com/writeas/writefreely/blob/develop/templates/chorus-collection-post.tmpl#L14) and [here](https://github.com/writeas/writefreely/blob/develop/templates/collection-post.tmpl#L15).

I opted to unescape `Content` in the `Summary` method because `Title` is already treated that way so this seemed more inline with the existing code. If you'd prefer updates to the templates instead, I can switch it up.

thebaer (approved):
Thanks for addressing this, @Antolius! I think the route you chose for this solution was the right one. Merging now!

Issue #374, opened 2020-08-26
Silenced users posts still show up when accessed directly


mrvdb:
The list of posts is correctly blocked, but individual posts are accessible it seems.

This is a partial screenshot of anonymous access to the url of the post. At the very least, the message is wrong.

![20200826_1258](https://user-images.githubusercontent.com/73067/91295940-1adfa400-e79c-11ea-8e47-f6fa47937758.png)

I haven't tested extensively with all silenced accounts. The 'user account URI' itself is properly not giving any content. (list of posts)

Expected behaviour:
- all users, including anonymous are not able to see any posts of a silenced user
- optionally, make an exception for the admin user, so the post content can be reviewed

thebaer:
Agreed, on this page we should check the currently authenticated user, and:

* If not logged in, return 404, just like on the silenced user's blog
* If logged in as `user` and `user.ID != post.OwnerID`, return 404
* If logged in as `user` and `!user.IsAdmin()`, return 404

Anyone should feel free to claim this task and open a pull request for a fix!

colin-axner:
hello! I'd like to try to tackle this issue

PR #384, opened 2020-09-17
fix accessibility of silenced user posts

colin-axner:
Change view post collection queries to verify that the authenticated user viewing a silenced collection is either the owner or admin. Previous behaviour was only checking if the post owner was also the collection owner which always returned true, causing the if statement to be skipped.

closes: #374 

---


mrvdb:
I've applied this patch for qua.name. Works for me with one exception, which may be a local issue, I haven't dived in yet.

Posts from silenced users are still visible in the 'reader' ( URI '/read' )  After restarting writefreely they are not visible anymore. I think posts from silenced users should also be removed from the 'reader' page without having to restart writefreely (possibly a caching issue?)

thebaer:
I agree, we should make it so silenced users are immediately removed from the Reader, too. This does sound like a caching issue -- maybe we could invalidate the cache anytime the Reader is enabled and a new user is silenced? Probably around here:

https://github.com/writeas/writefreely/blob/f534ee1deccbe5d161349554efa9aa0a2aac55df/read.go#L159-L161

colin-axner:
Sounds good. Would it be safe to set `tl.posts = nil` before calling `updateTimelineCache` to force the cache to be reset or alternatively pass in a boolean indicating a cache override?

thebaer:
I think the best way would be to pass a boolean that invalidates the cache. Since this is an expensive query, we prefer to keep `tl.posts` permanently populated after it first loads -- even if it serves stale content to a visitor or two, it keeps the Reader section responsive.

colin-axner:
maybe I am missing something, but wouldn't [`Memo.Invalidate`](https://github.com/writeas/web-core/blob/aacd7d3d0ce769028f9198eeaa13a9f1dcf5accd/memo/memo.go#L40) need to be modified to support this functionality. 

The cache is an unexported field so in order to forcibly reset the cache, `Invalidate` needs to take in a boolean to bypass the last duration. 

Should I instead add a third function `Reset` to forcibly reset the cache without breaking API?

thebaer:
> Should I instead add a third function Reset to forcibly reset the cache without breaking API?

Yes, I think that'd be the best route to go.

colin-axner:
made the changes based on the linked pr. I'm not sure if I reset the cache in the right place after silencing the user? It is unclear to me if there's a different handler called when an admin clicks "silence" on a user. 

Still need to manually test either via modifying the go.mod locally or merging the web-core pr and updating the commit

thebaer:
Moving this out of draft, now that writeas/web-core#11 is merged.

go.sum
```diff
@@ -1,13 +1,7 @@
 code.as/core/socks v1.0.0 h1:SPQXNp4SbEwjOAP9VzUahLHak8SDqy5n+9cm9tpjZOs=
 code.as/core/socks v1.0.0/go.mod h1:BAXBy5O9s2gmw6UxLqNJcVbWY7C/UPs+801CcSsfWOY=
-github.com/BurntSushi/toml v0.3.1 h1:WXkYYl6Yr3qBf1K79EBnL4mak0OimBfB0XUf9Vl28OQ=
 github.com/BurntSushi/toml v0.3.1/go.mod h1:xHWCNGjB5oqiDr8zfno3MHue2Ht5sIBksp03qcyfWMU=
 github.com/PuerkitoBio/goquery v1.5.1/go.mod h1:GsLWisAFVj4WgDibEWF4pvYnkVQBpKBKeU+7zCJoLcc=
```
colin-axner, line 4: I ran go mod tidy which is why there is a lot of deletions

go.mod
```diff
@@ -46,16 +40,11 @@ require (
 	github.com/writeas/nerds v1.0.0
 	github.com/writeas/saturday v1.7.2-0.20200427193424-392b95a03320
 	github.com/writeas/slug v1.2.0
-	github.com/writeas/web-core v1.2.1-0.20200813161734-68a680d1b03c
+	github.com/writeas/web-core v1.2.1-0.20210306215149-e05f572eb21f
```
colin-axner, line 43: I think `web-core` should have a tagged release with the recently merged change. But we can test the fix first

colin-axner (commented):
updated the web-core commit. Will try to test this week

go.sum
thebaer: Sounds good, thanks!

read.go
```diff
@@ -156,24 +156,31 @@ func viewLocalTimeline(app *App, w http.ResponseWriter, r *http.Request) error {
...
 }
 
-func updateTimelineCache(tl *localTimeline) {
-	// Fetch posts if enough time has passed since last cache
-	if tl.posts == nil || tl.m.Invalidate() {
+// updateTimelineCache will reset and update the cache if it is stale or
+// the boolean passed in is true.
+func updateTimelineCache(tl *localTimeline, reset bool) {
+	if reset {
+		tl.Reset()
```
thebaer, line 163: Getting this error when trying to compile:

```
../../read.go:163:5: tl.Reset undefined (type *localTimeline has no field or method Reset)
```

thebaer (changes requested):

read.go
colin-axner: Ahh yes, sorry about that

colin-axner:
I have tested these changes and they work as expected. When the user is silenced, the posts are not viewable directly or on the reader. When the user is unsilenced, the posts will be added back to the reader once the cache is refreshed

The UI is looking better. Looking forward to the next release

thebaer (approved):
Yep, tested and this works great! Thanks!

PR #389, opened 2020-09-25
remove unnecessary var in account.go

colin-axner:
`createdWithPass` is an unnecessary variable. The if statement that follows is unreachable since the hashed pass is set for the user and `HasPass` is set to true

---


thebaer (approved):
Thanks, @colin-axner! This looks good -- merging now.

PR #390, opened 2020-09-25
DRY get collection calls

colin-axner:
Replaced 4 repetitive if else with call to a helper function. In single user mode why don't the accounts just have an alias so you can query all modes with just `GetCollectionBy`? I wasn't sure where to place the helper function, but at the bottom of `app.go` seemed ok

---

### DIFF
--- a/activitypub.go
+++ b/activitypub.go
@@ -77,18 +77,11 @@ func handleFetchCollectionActivities(app *App, w http.ResponseWriter, r *http.Re
 	vars := mux.Vars(r)
 	alias := vars["alias"]
 
-	// TODO: enforce visibility
-	// Get base Collection data
-	var c *Collection
-	var err error
-	if app.cfg.App.SingleUser {
-		c, err = app.db.GetCollectionByID(1)
-	} else {
-		c, err = app.db.GetCollection(alias)
-	}
+	c, err := GetCollection(app, alias)
 	if err != nil {
 		return err
 	}
+
 	silenced, err := app.db.IsUserSilenced(c.OwnerID)
 	if err != nil {
 		log.Error("fetch collection activities: %v", err)
@@ -111,18 +104,11 @@ func handleFetchCollectionOutbox(app *App, w http.ResponseWriter, r *http.Reques
 	vars := mux.Vars(r)
 	alias := vars["alias"]
 
-	// TODO: enforce visibility
-	// Get base Collection data
-	var c *Collection
-	var err error
-	if app.cfg.App.SingleUser {
-		c, err = app.db.GetCollectionByID(1)
-	} else {
-		c, err = app.db.GetCollection(alias)
-	}
+	c, err := GetCollection(app, alias)
 	if err != nil {
 		return err
 	}
+
 	silenced, err := app.db.IsUserSilenced(c.OwnerID)
 	if err != nil {
 		log.Error("fetch collection outbox: %v", err)
@@ -174,18 +160,11 @@ func handleFetchCollectionFollowers(app *App, w http.ResponseWriter, r *http.Req
 	vars := mux.Vars(r)
 	alias := vars["alias"]
 
-	// TODO: enforce visibility
-	// Get base Collection data
-	var c *Collection
-	var err error
-	if app.cfg.App.SingleUser {
-		c, err = app.db.GetCollectionByID(1)
-	} else {
-		c, err = app.db.GetCollection(alias)
-	}
+	c, err := GetCollection(app, alias)
 	if err != nil {
 		return err
 	}
+
 	silenced, err := app.db.IsUserSilenced(c.OwnerID)
 	if err != nil {
 		log.Error("fetch collection followers: %v", err)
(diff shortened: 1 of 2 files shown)


activitypub.go
```diff
@@ -77,18 +77,11 @@ func handleFetchCollectionActivities(app *App, w http.ResponseWriter, r *http.Re
 	vars := mux.Vars(r)
 	alias := vars["alias"]
 
-	// TODO: enforce visibility
-	// Get base Collection data
```
thebaer, line 81: We should retain this TODO comment, as this PR doesn't solve it.

Also, it probably wouldn't hurt to keep this informative comment below the TODO, either (unless you have a particular reason for removing it).

activitypub.go
```diff
@@ -111,18 +104,11 @@ func handleFetchCollectionOutbox(app *App, w http.ResponseWriter, r *http.Reques
...
```
thebaer, line 115: See above

activitypub.go
```diff
@@ -174,18 +160,11 @@ func handleFetchCollectionFollowers(app *App, w http.ResponseWriter, r *http.Req
...
```
thebaer, line 178: See above

activitypub.go
```diff
@@ -229,18 +208,11 @@ func handleFetchCollectionFollowing(app *App, w http.ResponseWriter, r *http.Req
...
```
thebaer, line 233: See above

app.go
```diff
@@ -902,3 +902,13 @@ func ServerUserAgent(hostName string) string {
 	}
 	return "Go (" + serverSoftware + "/" + softwareVer + hostUAStr + ")"
 }
+
+// GetCollection returns the collection associated with the alias. If the application is in
+// single user mode then the collection associated with ID 1 (admin) is returned.
```
thebaer, line 907: As a first-time reader, the most confusing things about this func are:

* The app config's effects on the return values (that it affects the results at all)
* Conditions when the `alias` param is actually used

So I think the documentation could be improved by specifying that: The `alias` is only utilized in multi-user mode, otherwise the first **collection** in the database is returned (not necessarily the collection associated with a particular user). You might also mention that an empty `alias` value is acceptable.

Throughout the app, there are many different scenarios where we want to get a collection from the database. So I think this func name should be more descriptive, and explain _when_ someone should use it, e.g.:

* `GetCollectionFromURL`
* `GetCollectionFromAlias`
* `GetCollectionForView`
* etc.

thebaer:
> In single user mode why don't the accounts just have an alias so you can query all modes with just `GetCollectionBy`?

In single-user mode, no alias is given in the URL -- e.g. `https://matt.writefreely.dev/` -- so we can't look up the collection from the URL alone. This is compared to multi-user mode, where the alias is in the URL, e.g. `https://write.as/matt/`, (`matt` is the alias).

thebaer (changes requested):
There's always a balance between readability and maintainability, right? I appreciate that this removes repeated code, but it also shifts reading from code logic to func names and documentation. So if we're going to make this change, we should make sure those are correct and easy to understand.

activitypub.go
colin-axner: sorry, this was accidental. Will add it back. 

Is there anything I can add to the "enforce visibility" to make it more descriptive? It is unclear to me what it implies. Maybe I can link an issue?

app.go
colin-axner: Thanks for the feedback. 

If the function needs to get a collection and getting this collection depends on whether there is or isn't a alias in the URL, then I think it'd be good to put all this logic in one place. 

> In single-user mode, no alias is given in the URL -- e.g. https://matt.writefreely.dev/ -- so we can't look up the collection from the URL alone. This is compared to multi-user mode, where the alias is in the URL, e.g. https://write.as/matt/, (matt is the alias).

I think `GetCollectionFromURL` would make a lot of sense. We could continue using the logic as rewritten but in all cases if the config is in single user mode then the `alias == ""` right? The function signature could be modified to take in the request and retrieve the alias from the URL/check the app for its user mode. 

Is the alias typically used for anything else other than getting the collection?

activitypub.go
```diff
@@ -133,12 +120,6 @@ func handleFetchCollectionOutbox(app *App, w http.ResponseWriter, r *http.Reques
 	}
 	c.hostName = app.cfg.App.Host
 
-	if app.cfg.App.SingleUser {
```
colin-axner, line 136: I hesitated to remove this, but IMO if the `alias != c.Alias` this should be a bug. It seems in this function, because it is using the activity pub outbox, the alias of the collection must be set so I think it is safe to treat this how multi user applications treat collections in other areas of the code

collections.go
```diff
@@ -457,12 +457,8 @@ func fetchCollection(app *App, w http.ResponseWriter, r *http.Request) error {
 		return handleFetchCollectionActivities(app, w, r)
 	}
 
-	vars := mux.Vars(r)
-	alias := vars["alias"]
-
-	// TODO: move this logic into a common getCollection function
```
colin-axner, line 463: I think this TODO has been addressed

colin-axner (commented):
I changed `GetCollection` to `GetCollectionFromAliasVar` which I think is the most descriptive name I could think of. It collides slightly with `collectionAliasFromReq` which does different logic. The function I added gets the alias from the variable labeled "alias" so I think the name should be fine.

Issue #305, opened 2020-04-27
Post titled "Tags" cannot be edited


cjeller1592:
### Describe the bug
When you create a post titled "Tags" & go to edit the page, you get a 404: "This Page Is Missing. Are you sure it was ever here?".

### Expected behavior
Should be able to go in and edit the post.

**Version or last commit**:
v0.12

Crocmagnon:
An easy fix would maybe to prevent using « tags » as a slug? In the slug generation function, check if « tags » is used then just add a dash or something.
If the user tried to use « tags », prevent it and display a helpful error message.

thebaer:
This bug is caused by this route: https://github.com/writeas/writefreely/blob/24fa3d6863ab6a3047c8bbe75203efb09ce7b894/routes.go#L207

The application assumes "edit" is the tag, and since it doesn't find any posts with `#edit` in them, it fails. Here are two ways I could think of to fix this:

* Drop this route entirely, since we don't officially use it anyway, or
* Add a special "/tags/edit" route here that uses the correct handler function

Any thoughts?

cjeller1592:
If the route isn't officially used anyway, I'd say drop it entirely. Would there be any other repercussions to removing that route that you can think of @thebaer?

thebaer:
Nope, we can drop it.

PR #397, opened 2020-10-03
Drop the /tags/{tag} route

mrvdb:
fixes issue #305

Just drops the route as suggested.


thebaer (approved):
Thanks for the fix, @mrvdb! Merging now.

PR #405, opened 2020-10-22
Generic OAuth userinfo properies now configurable

dariusk:
When connecting to a generic OAuth provider, you are never sure what object you'll be receiving from the userinfo endpoint (it isn't actually specified anywhere). So this commit adds mapping values to the generic oauth configuration section of config.ini, allowing the user to specify which keys in the remote endpoint it expects to read the UserID, Username, DisplayName, and Email from. Default values if unspecified remain as they were before this commit.

---

### DIFF
--- a/oauth_generic.go
+++ b/oauth_generic.go
@@ -16,6 +16,10 @@ type genericOauthClient struct {
 	InspectLocation  string
 	CallbackLocation string
 	Scope            string
+  MapUserID        string
+  MapUsername      string
+  MapDisplayName   string
+  MapEmail         string
 	HttpClient       HttpClient
 }
 
@@ -104,12 +108,28 @@ func (c genericOauthClient) inspectOauthAccessToken(ctx context.Context, accessT
 		return nil, errors.New("unable to inspect access token")
 	}
 
-	var inspectResponse InspectResponse
-	if err := limitedJsonUnmarshal(resp.Body, infoRequestMaxLen, &inspectResponse); err != nil {
+	// since we don't know what the JSON from the server will look like, we create a
+	// generic interface and then map manually to values set in the config
+	var genericInterface interface{}
+	if err := limitedJsonUnmarshal(resp.Body, infoRequestMaxLen, &genericInterface); err != nil {
 		return nil, err
 	}
-	if inspectResponse.Error != "" {
-		return nil, errors.New(inspectResponse.Error)
+
+	m := genericInterface.(map[string]interface{})
+
+	// map each relevant field in inspectResponse to the mapped field from the config
+	var inspectResponse InspectResponse
+	if (m[c.MapUserID] != nil) {
+	  inspectResponse.UserID = m[c.MapUserID].(string)
+	}
+	if (m[c.MapUsername] != nil) {
+	  inspectResponse.Username = m[c.MapUsername].(string)
+	}
+	if (m[c.MapDisplayName] != nil) {
+	  inspectResponse.DisplayName = m[c.MapDisplayName].(string)
+	}
+	if (m[c.MapEmail] != nil) {
+	  inspectResponse.Email = m[c.MapEmail].(string)
 	}
 
 	return &inspectResponse, nil
--- a/oauth_generic.go
+++ b/oauth_generic.go
@@ -16,10 +16,10 @@ type genericOauthClient struct {
 	InspectLocation  string
 	CallbackLocation string
 	Scope            string
-  MapUserID        string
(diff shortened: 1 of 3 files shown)


oauth_generic.go
```diff
@@ -104,12 +108,28 @@ func (c genericOauthClient) inspectOauthAccessToken(ctx context.Context, accessT
...
+	var genericInterface interface{}
+	if err := limitedJsonUnmarshal(resp.Body, infoRequestMaxLen, &genericInterface); err != nil {
 		return nil, err
 	}
-	if inspectResponse.Error != "" {
-		return nil, errors.New(inspectResponse.Error)
+
+	m := genericInterface.(map[string]interface{})
+
+	// map each relevant field in inspectResponse to the mapped field from the config
```
dariusk, line 120: I feel like there must be a better way to do this than 4 if statements in a row but it was all I could figure out.

dariusk:
I am _not_ a golang programmer and probably made some very basic mistakes here. [This in particular, the 4 if statements in a row](git@github.com:dariusk/writefreely.git), seems like it could be improved.

oauth_generic.go
```diff
@@ -104,12 +108,28 @@ func (c genericOauthClient) inspectOauthAccessToken(ctx context.Context, accessT
 		return nil, errors.New("unable to inspect access token")
 	}
 
-	var inspectResponse InspectResponse
-	if err := limitedJsonUnmarshal(resp.Body, infoRequestMaxLen, &inspectResponse); err != nil {
+	// since we don't know what the JSON from the server will look like, we create a
+	// generic interface and then map manually to values set in the config
+	var genericInterface interface{}
```
thebaer, line 113: Maybe use a `map[string]interface{}` as the type here, instead of an `interface{}`? That should reduce some code and get rid of the following `m` variable below.

oauth_generic.go
```diff
@@ -104,12 +108,28 @@ func (c genericOauthClient) inspectOauthAccessToken(ctx context.Context, accessT
...
 	}
-	if inspectResponse.Error != "" {
-		return nil, errors.New(inspectResponse.Error)
+
+	m := genericInterface.(map[string]interface{})
+
+	// map each relevant field in inspectResponse to the mapped field from the config
+	var inspectResponse InspectResponse
+	if (m[c.MapUserID] != nil) {
+	  inspectResponse.UserID = m[c.MapUserID].(string)
```
thebaer, line 123: Currently, this will crash if the value isn't a `string`. Instead, you'll want to also check that the type coercion worked by checking for a second return value:

```go
var ok bool
inspectResponse.UserID, ok = m[c.MapUserID].(string)
```

However, since we only want to avoid crashes and fall back to the default value if the type coercion fails, you can just ignore the second value:

```go
inspectResponse.UserID, _ = m[c.MapUserID].(string)
```

([Some demo code with more info about this](https://play.golang.org/p/7tZe4Gpc0ZX))

oauth_generic.go
```diff
@@ -104,12 +108,28 @@ func (c genericOauthClient) inspectOauthAccessToken(ctx context.Context, accessT
...
+	}
+	if (m[c.MapUsername] != nil) {
+	  inspectResponse.Username = m[c.MapUsername].(string)
```
thebaer, line 126: See above for this and remaining values.

thebaer (changes requested):
It's all good -- this is almost there! Parsing can be a bit tricky, but there are just a few changes to make here. Let me know if you have any questions about this stuff.

Also, just need to run `go fmt` on the file so formatting is consistent.

dariusk:
Okay, updated. (The more I learn about go, the more I like it.)

dariusk:
Oops, one more requested change incoming.

dariusk:
Okay, did a force push so now all requested changes are in the second commit.

oauth_generic.go
dariusk: Done in b262fa144c995d206c02aa4ed378c9d0b234c77a

oauth_generic.go
dariusk: Done in b262fa144c995d206c02aa4ed378c9d0b234c77a

oauth_generic.go
dariusk: Done in b262fa144c995d206c02aa4ed378c9d0b234c77a

thebaer (approved):
Finally getting back to this -- sorry for the delay. This looks good! We'll get it merged.

Issue #401, opened 2020-10-09
go tests failing on develop


colin-axner:
on latest develop, if I run:

```go
go test
ERROR: 2020/10/09 16:43:21 oauth.go:148: viewOauthInit error: pretend unable to write state error
--- FAIL: TestViewOauthCallback (0.00s)
    --- FAIL: TestViewOauthCallback/success (0.00s)
        oauth_test.go:248: 
            	Error Trace:	oauth_test.go:248
            	Error:      	Received unexpected error:
            	            	/login
            	Test:       	TestViewOauthCallback/success
        oauth_test.go:249: 
            	Error Trace:	oauth_test.go:249
            	Error:      	Not equal: 
            	            	expected: 307
            	            	actual  : 200
            	Test:       	TestViewOauthCallback/success
2020/10/09 16:43:21 Loading testdata/config.ini configuration...
--- FAIL: TestUpdatesRoundTrip (1.09s)
    --- FAIL: TestUpdatesRoundTrip/Release_URL (0.00s)
        updates_test.go:36: Malformed Release URL: https://blog.writefreely.org/version-
FAIL
exit status 1
FAIL	github.com/writeas/writefreely	1.104s
```

tokudan:
fails on 0.12 as well, not sure if it's related, as the error is different, but the location is similar.
```
running tests
=== RUN   TestUnmarshalActor
--- PASS: TestUnmarshalActor (0.00s)
=== RUN   TestOAuthDatastore
    database_test.go:12: skipping mysql tests
--- SKIP: TestOAuthDatastore (0.00s)
=== RUN   TestViewOauthInit
=== RUN   TestViewOauthInit/success
=== RUN   TestViewOauthInit/state_failure
ERROR: 2020/10/09 23:56:51 oauth.go:140: viewOauthInit error: pretend unable to write state error
--- PASS: TestViewOauthInit (0.00s)
    --- PASS: TestViewOauthInit/success (0.00s)
    --- PASS: TestViewOauthInit/state_failure (0.00s)
=== RUN   TestViewOauthCallback
=== RUN   TestViewOauthCallback/success
--- FAIL: TestViewOauthCallback (0.00s)
    --- FAIL: TestViewOauthCallback/success (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
        panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x95e3f1]

[...]

PR #406, opened 2020-10-26
fix oauth tests

colin-axner:
Fixes the first two tests in #401 .

I noticed the test was failing [here] so I updated the app config to be set to Open Registration.

Then the test began panicing because the templates weren't loaded so I added InitTemplates to load them. The test case is already labeled `success` so I assume the written code should be OK.

```
go test
ERROR: 2020/10/26 13:34:29 oauth.go:148: viewOauthInit error: pretend unable to write state error
2020/10/26 13:34:29 Loading templates...
2020/10/26 13:34:29 Loading pages...
2020/10/26 13:34:29 Loading user pages...
2020/10/26 13:34:29 Loading testdata/config.ini configuration...
--- FAIL: TestUpdatesRoundTrip (1.09s)
    --- FAIL: TestUpdatesRoundTrip/Release_URL (0.00s)
        updates_test.go:36: Malformed Release URL: https://blog.writefreely.org/version-
FAIL
exit status 1
FAIL	github.com/writeas/writefreely	1.169s
``` 

updated output. The first 5 lines are just logs from running the tests. Is there a way to disable the logs from being outputted in the console?

---

### DIFF
--- a/oauth_test.go
+++ b/oauth_test.go
@@ -207,8 +207,14 @@ func TestViewOauthInit(t *testing.T) {
 func TestViewOauthCallback(t *testing.T) {
 	t.Run("success", func(t *testing.T) {
 		app := &MockOAuthDatastoreProvider{}
+
+		// create a compatibile configuration for testing
+		config := app.Config()
+		config.App.SingleUser = false
+		config.App.OpenRegistration = true
+
 		h := oauthHandler{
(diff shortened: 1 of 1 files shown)


cjeller1592:
I got the same output from running tests on this branch — thanks @colin-axner! Wonder if @thebaer has any input on how to disable the logs.

PR #418, opened 2020-12-15
Use HTMLContent field in RSS feed

thebaer:
This re-uses the HTMLContent feed for the full HTML content in the RSS feed, instead of again generating HTML from Markdown.

This keeps things more consistent throughout the application and reduces work when rendering the feed.

---


cjeller1592:
This is the `content:encoded` element in the feed right? PR looks good on my end!

thebaer:
That's right -- it should overall work the exact same. Thanks for reviewing!

thebaer:
Merging now!

PR #425, opened 2021-01-09
Use newer mixin syntax for less 4.x compatibility

mrvdb:
See https://github.com/less/less.js/releases/tag/v4.0.0

This syntax also works with less 3.x


mrvdb:
Anything needed for this, seems non intrusive and mergeable?

mrvdb:
Irrelevant now, see https://github.com/less/less.js/releases/tag/v4.1.0

PR #426, opened 2021-01-14
Fixes problem where all links were breaking in gopher

mostfunkyduck:
The gopher integration was not setting Host and Port properly in item structs.
This will derive the host from the configured host by stripping the protocol from the URI and will set the port based on configuration.

---

- [X ] I have signed the [CLA](https://phabricator.write.as/L1)

### DIFF
--- a/gopher.go
+++ b/gopher.go
@@ -14,6 +14,7 @@ import (
 	"bytes"
 	"fmt"
 	"io"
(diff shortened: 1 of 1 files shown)


thebaer (approved):
Thanks for finding and fixing this! Looks good to me, merging now.

PR #429, opened 2021-01-24
Case insensitive language highlighting matching

x4e:
This automatically lowercases language names used in code blocks when finding highlighting scripts for them, since highlightJS defines all languages with lowercase names.

This for example allows using `C++` as a language name, which previously would not have worked.

- [Y] I have signed the [CLA](https://phabricator.write.as/L1)


thebaer (approved):
Looks good to me! Thanks for fixing this.

PR #431, opened 2021-01-27
Stop private and protected blogs from federating

### LINKED ISSUE #403: Posts still federate after setting blog to "Private" or "Password-protected"
### Observed behavior
I am running an instance with federation enabled for posts. If someone has a blog that is Unlisted, and someone subscribes to that blog via ActivityPub, and then they set the blog to Private, posts that are made from then on will still be federated to other people. 

This is also true for "Password-Protected". The full text of the article federates out to any subscribers.

### Expected behavior
I would expect that federation would halt while Private, similar to how the RSS feed returns a 404 after you switch to Private.

### Application configuration
 - Multi-user mode
 - sqlite
 - Open registration
 - Federation enabled
 - Disabled http auth
 - Multiple blogs per user

**Version or last commit**: latest `develop` as of 3493921837b18cc4239bd3d8252d71f97d5e1d5b

Dak425:
Fixes  #403 

---

- [ x] I have signed the [CLA](https://phabricator.write.as/L1)

### DIFF
--- a/activitypub.go
+++ b/activitypub.go
@@ -624,14 +624,6 @@ func deleteFederatedPost(app *App, p *PublicPost, collID int64) error {
 }
 
 func federatePost(app *App, p *PublicPost, collID int64, isUpdate bool) error {
-	if debugging {
-		if isUpdate {
-			log.Info("Federating updated post!")
-		} else {
-			log.Info("Federating new post!")
-		}
-	}
-
 	// If app is private, do not federate
 	if app.cfg.App.Private {
 		return nil
@@ -640,6 +632,14 @@ func federatePost(app *App, p *PublicPost, collID int64, isUpdate bool) error {
 	// Do not federate posts from private or protected blogs
 	if p.Collection.Visibility == CollPrivate || p.Collection.Visibility == CollProtected {
 		return nil
+	}
+
+	if debugging {
+		if isUpdate {
+			log.Info("Federating updated post!")
+		} else {
+			log.Info("Federating new post!")
+		}
 	}
 
 	actor := p.Collection.PersonObject(collID)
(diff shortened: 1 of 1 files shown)


Dak425:
@thebaer @robjloranger Pull request needs reviewed and merged

thebaer (changes requested):
This looks good! Only one change I'd suggest: let's move this logic before the `log.Info()` calls above, so debug messages are correct.

thebaer:
With these changes, we can also clean up / de-duplicate some logic around the various `federatePost()` calls throughout the app, like here: https://github.com/writeas/writefreely/blob/2c1d3a51af8a0bd0dc11d86a7e0e4fab46b672ff/posts.go#L619-L621

but I'd say that's optional to include in this PR. You can make those changes here if you'd like, but otherwise we'll get this merged after the previously mentioned changes are made.

Dak425:
> With these changes, we can also clean up / de-duplicate some logic around the various `federatePost()` calls throughout the app, like here: https://github.com/writeas/writefreely/blob/2c1d3a51af8a0bd0dc11d86a7e0e4fab46b672ff/posts.go#L619-L621
> 
> but I'd say that's optional to include in this PR. You can make those changes here if you'd like, but otherwise we'll get this merged after the previously mentioned changes are made.

I think cleaning up the redundant logic can be its own issue. If anything I'll do that one as well.

Dak425:
> This looks good! Only one change I'd suggest: let's move this logic before the `log.Info()` calls above, so debug messages are correct.

Ah true that makes more sense, change has been made.

thebaer:
> I think cleaning up the redundant logic can be its own issue. If anything I'll do that one as well.

Sounds great!

thebaer (approved):
Looks good to me, thanks!

PR #435, opened 2021-02-10
Add conditional for preserving lang metadata

### LINKED ISSUE #280: Language metadata gets reset to English after editing a post
Brought up originally on [the forum](https://discuss.write.as/t/language-metadata-gets-reset-to-english-after-re-publishing-a-post/1183):

>Normally, after setting a language for a post, it should stick and should not change unless the author modifies it again. However if you edit the post and (re)publish it again it causes the language field to reset back to English.

### Steps to reproduce
1. Go to a post and click 'Edit.' Go into the metadata and change the Language to something else and save your changes. Publish the post.
2. Now edit the post again, this time the body of the post, and publish.
3. Go back to the metadata of the post. Note how it is now 'English.'

### Expected behavior
Post should keep the metadata language after edited, not revert to English.

cjeller1592:
This PR fixes #280 — if a post is updated, it will retain the language metadata rather than reset back to English.

---

### DIFF
--- a/templates/pad.tmpl
+++ b/templates/pad.tmpl
@@ -194,8 +194,13 @@
 				body: post.content,
 				title: post.title,
 				font: font,
-				lang: lang
-			};
+			{{ if or .Post.Slug .Post.Id }}
+                        };
+                        {{ else }}
+                                lang: lang
+                        };
+			{{ end }}
+			
 			{{ if .Post.Slug }}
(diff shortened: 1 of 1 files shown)


thebaer (approved):
This fixes it, thanks @cjeller1592! I think we could clean up the style a bit to prevent errors in the future, but I think it's good to merge now.

PR #445, opened 2021-03-30
Remove writeas/nerds/store pkg dependency

thebaer:
This moves a few needed funcs in WriteFreely from `github.com/writeas/nerds` to `github.com/writeas/web-core` (a package we already use), reducing the number of external dependencies.

---


thebaer:
Merging now.

PR #452, opened 2021-04-07
Disable Ace editor on touchscreen devices

thebaer:
As [reported on our forum](https://discuss.write.as/t/custom-css-using-safari-on-ipad/1720), Ace doesn't work well with touchscreen devices. To fix this, we fall back to a plain textarea on touchscreen devices.

---


thebaer:
[Confirmed via forum](https://discuss.write.as/t/classic-editor-v2/2625/22) that this is fixed on Write.as instance. Merging now.

PR #453, opened 2021-04-08
Use Go 1.16 asset embedding

SamWhited:
This removes the need for go-bindata using Go's new built-in embedding. I don't know how many versions of Go you support, so it may be a touch early for this, but I wanted to go ahead and float the idea to simplify the build.


SamWhited:
Sorry, I think I've done this before but I forgot you required a CLA.

PR #457, opened 2021-04-26
Support international domain names

thebaer:
This internally converts the configured host name into its Punycode ASCII representation, while showing users the correct Unicode domain name. It enables admins to host a WriteFreely instance at a domain like https://bär.writefreely.dev (Punycode: https://xn--br-via.writefreely.dev).

Tested and verified this works in the fediverse. Demo post: https://xn--br-via.writefreely.dev/testing-non-ascii-domains

---


thebaer:
Merging now.

PR #464, opened 2021-05-25
Final API changes

thebaer:
This makes a few API changes first implemented on our Write.as instance to support a large multi-user migration. Changes:

* *Publish a Post* endpoint: return a `url` property of the final post URL
* Support updating collections only with a `monetization_pointer` field supplied
* Support publishing posts with only a title

---


thebaer:
Merging now

PR #471, opened 2021-06-22
Don't render title as list item

### LINKED ISSUE #470: Date in Title
### Describe the bug
If you enter a date in the title `12. april` then it becomes `1. april`
in the database it is correct
but if you write `12.april` it works without problems

1. Create a new post
2. enter a date in the title
3. save the post

### Expected behavior
the correct date should be displayed

### Application configuration
 - **Single mode or Multi-user mode?** Multi-user mode
 - **Database?** mysql
 - **Open registration?** no
 - **Federation enabled?** yes

**Version or last commit**: 0.13.0

thebaer:
This fixes an issue where "12. April" would get rendered as "1. April" because it looks like a Markdown list item to our renderer. Now, we parse titles as titles, instead of standalone text, which causes the renderer to give us the results we want. This also adds some basic tests for the `applyBasicMarkdown()` func.

Closes #470

---

### DIFF
--- a/postrender.go
+++ b/postrender.go
@@ -11,6 +11,7 @@
 package writefreely
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
 	"html"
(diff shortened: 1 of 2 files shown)


thebaer:
Tested on Write.as and found a couple issues, patched with those last few commits. Merging now.

PR #474, opened 2021-06-23
Error when generic OAuth provider doesn't return a user ID

### LINKED ISSUE #469: [Security] Generic OAuth allows anyone to authenticate as a user if map_user_id is not configured correctly
### Describe the bug

I was attempting to configure Generic OAuth for my WriteFreely instance with my Mastodon instance. Initially, I followed what's described on https://writefreely.org/docs/latest/admin/config, and everything seemed to be working. I could link my Mastodon account on WriteFreely, and I could log into my WriteFreely account with my Mastodon account just fine... until I noticed that everybody that has an account on my Mastodon instance now authenticates as me on the WriteFreely instance.

A close inspection on the database revealed that the `remote_user_id` field in the database was empty (null). This resulted in the single record being matched for every OAuth login request, thus resulting the aforementioned behavior.

After digging around in the source code, I realized that Mastodon does not return the user id in the field `user_id`, but instead in `id`, and WriteFreely needs a special configuration `map_user_id` for that (and also a few other `map_*` options). These are mentioned nowhere in the documentation.

So basically, the issue here is:

1. The OAuth code **should not** appear as being working fine when in fact being horribly broken (i.e. when `remote_user_id` cannot be fetched). There should be some big red error / warning in the logs to notify the administrator of the situation, instead of having to dig around in the source code to find the solution.
2. The `map_*` options should be included somewhere noticeable in the documentation.

1. Set up OAuth authentication with Mastodon without providing `map_user_id`
2. Link any Mastodon account with any WriteFreely account 
3. Notice that now any Mastodon account registered on the linked instance can authenticate with said WriteFreely account

### Expected behavior

1. Step 2 in "Steps to reproduce" should have failed in the first place.
2. Step 3 in "Steps to reproduce" should also fail (it should not match records with `remote_user_id` being null)

**Version or last commit**: 0.13.0

thebaer:
This returns an error and privately (via logs) prompts the admin to add a `map_user_id` config value.

Fixes #469

---

### DIFF
--- a/oauth_generic.go
+++ b/oauth_generic.go
@@ -1,8 +1,20 @@
+/*
+ * Copyright © 2020-2021 A Bunch Tell LLC and respective authors.
+ *
+ * This file is part of WriteFreely.
+ *
+ * WriteFreely is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, included
+ * in the LICENSE file in this source code package.
+ */
+
 package writefreely
 
 import (
 	"context"
 	"errors"
+	"fmt"
+	"github.com/writeas/web-core/log"
 	"net/http"
 	"net/url"
 	"strings"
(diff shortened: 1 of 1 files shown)


thebaer:
With documentation changes merged in writefreely/documentation#35, this is ready to merge too.

PR #475, opened 2021-06-23
Remove additional slash in /feed/ redirect

thebaer:
This fixes an issue with blogs redirecting to `https://instance.tld/user//feed/` (with an additional slash), as reported [on the forum](https://discuss.write.as/t/rss-feeds-are-broken/2949).

---


thebaer:
Merging now.

PR #477, opened 2021-06-25
Fix SQLite database lock on Reader

### LINKED ISSUE #467: After updating to v0.13.0 /read will not load
Since updating the reader feed does not load. My reverse proxy sends a 504 and loads just hang on the debug message 
```[READ] Updating post cache```

- Multi-user mode
- sqlite
- Open registration? no
- Federation enabled? yes

thebaer:
Fetching posts for the Reader involves an additional query on each row, which previously ran into our connection limit to the database and caused it to lock up. This increases the connection limit from 1 to 2, to allow this.

This is meant to be a quick, safe fix, but there could always be a better solution.

Fixes #467

---


thebaer:
Merging now.

PR #482, opened 2021-06-27
Remove rel=nofollow attribute in certain situations

### LINKED ISSUE #314: Links have rel="nofollow"
### Describe the bug
Links added by the user on an article or a pinned page have a `rel="nofollow"` attribute.

Add a link in an article, either with markdown or HTML.
If the link is HTML and the user explicitly specifies another value for the `rel` attribute, this value is overridden when rendering the page.

### Expected behavior
I see two issues with this:

* This prevents Mastodon profile validation (which requires a `rel="me"`)
* This may reduce the visibility of the blogs since search engines apparently respect this directive

From the MDN:
> Indicates that the linked document is not endorsed by the author of this one, for example if it has no control over it, if it is a bad example or if there is commercial relationship between the two (sold link).
- https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types

IMHO this:
* shouldn't be applied at all on internal links (links to `href="#something"` or to other pages of the blog)
* should be applied by default on external links if the user doesn't specify anything else via a custom HTML anchor.

### Application configuration
 - **Single mode or Multi-user mode?**  single
 - **Database?** sqlite
 - **Open registration?** no
 - **Federation enabled?** yes

**Version or last commit**: 037fc40

thebaer:
Previously, WriteFreely always applied `rel="nofollow"` on all blog links, since posts on multi-user instances are published by untrusted users, and this helps protect such instances' reputation with search engines. This change alters that behavior by _not_ applying `rel="nofollow"` if the instance is set to single-user mode, since we can assume that all content is trusted, then. When merged, it will close #314.

Still to do:

- [ ] Allow user to manually add `rel` attributes on HTML links

Still up for discussion:

* Should we use any other signals to determine whether or not to disable `rel="nofollow"`?
* Is there a reason we might add an explicit configuration option for this? Or could we leverage another one for that?

---

### DIFF
--- a/postrender.go
+++ b/postrender.go
@@ -119,7 +119,7 @@ func (p *PublicPost) augmentReadingDestination() {
 }
 
 func applyMarkdown(data []byte, baseURL string, cfg *config.Config) string {
-	return applyMarkdownSpecial(data, false, baseURL, cfg)
+	return applyMarkdownSpecial(data, baseURL, cfg, false)
 }
 
 func disableYoutubeAutoplay(outHTML string) string {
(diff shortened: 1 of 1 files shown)


thebaer:
Merging now.

PR #483, opened 2021-06-27
Log user out when authenticated as deleted user

thebaer:
Now when we check for the user at certain times and find that the user doesn't exist in the database, we log them out and send them back to the home page.

Previously, this could happen when e.g. the user is logged in on multiple browsers and then deletes their account -- they'll be logged out in the browser they deleted from, but not from other browser sessions.

---


thebaer:
Working well on Write.as, merging now.

PR #498, opened 2021-07-22
Fix import path for go-gopher

prologic:
Project has moved to [git.mills.io/prologic/go-gopher](https://git.mills.io/prologic/go-gopher).

For details on why see: https://github.com/prologic

Thank you! 🙇


thebaer:
Thanks for submitting this, @prologic! I actually addressed this in #497, and had to also update the package version in `go.mod` to get things to build correctly. I appreciate the contribution and would've merged this instead, but since the work is already done and tested there, I think we'll go with that one.

prologic:
> Thanks for submitting this, @prologic! I actually addressed this in #497, and had to also update the package version in `go.mod` to get things to build correctly. I appreciate the contribution and would've merged this instead, but since the work is already done and tested there, I think we'll go with that one.

No worries mate! Thanks 👌

PR #501, opened 2021-08-05
Fix Gopher collections query

mnlg:
Accessing to the Writefreely instance via Gopher throws this error:

```
ERROR: 2021/08/05 13:47:47 database.go:1693: Failed selecting public collections: ambiguous column name: id
ERROR: 2021/08/05 13:47:47 handle.go:990: failed: Couldn't retrieve public collections.
```

Instead of changing it to ORDER BY c.id it seems to make more sense ordering the collection by title instead.

---


thebaer:
Thanks for submitting this fix!

At first, I wasn't sure if it made the most sense to change what we're actually sorting by. But I agree, sorting by title makes a lot of sense. Merging now!

PR #503, opened 2021-08-13
Fix date format in anonymous posts

mnlg:
The date format for anonymous posts is always displayed in the default English format.

The anonymous posts SQL query doesn't include the language field which is then used to format the date. As the language field is empty it defaults to the English format.

```
	loc := monday.FuzzyLocale(p.Language.String)
	res.DisplayDate = monday.Format(p.Created, monday.LongFormatsByLocale[loc], loc)
```

Adding the language to the anonymous posts query fixes the issue.

---


thebaer:
Thanks for finding and fixing this! Will verify and merge this soon.

thebaer:
Thanks again for this. Looks good! Merging now.

thebaer (approved):

PR #505, opened 2021-08-21
Templates: use relative URLs for static assets

lt3Dev:
I noticed most asset links use relative URLs, except for a few. This commit remedies this inconsistency. OpenGraph embeds were not changed, because in my experience, some embed scrapers require absolute URLs.

Motivation: The site I use has an onion and a clearnet site, and currently, visiting the onion loads fonts, styles, and the favicon
over the clearnet.

---

### DIFF
--- a/templates/base.tmpl
+++ b/templates/base.tmpl
@@ -2,8 +2,8 @@
 <html>
 	<head>
 		{{ template "head" . }}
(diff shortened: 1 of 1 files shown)


thebaer (approved):
Thanks again for this! Merging now.

PR #511, opened 2021-09-28
Hotfix/accessibility

mnlg:
This PR fixes a couple of accessibility issues reported by Web Vitals. 

**Some elements have a [tabindex] value greater than 0**

> Although technically valid, using a tabindex greater than 0 is considered an anti-pattern because it shifts the affected element to the end of the tab order. This unexpected behavior can make it seem like some elements can't be accessed via keyboard, which is frustrating for users who rely on assistive technologies.

**<dl>s do not contain only properly ordered <dt> and <dd> groups, <script>, or <template> elements** and **Definition list items are not wrapped in <dl> elements**

> Screen readers and other assistive technologies have a specific way of announcing definition lists. When definition lists are not properly marked up, assistive technologies may produce confusing or inaccurate output.

---

### DIFF
--- a/pages/landing.tmpl
+++ b/pages/landing.tmpl
@@ -88,17 +88,17 @@ form dd {
 					<label>
 						<dt>Username</dt>
 						<dd>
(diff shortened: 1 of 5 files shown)


thebaer:
Thank you for submitting this! I'll take a look soon.

Issue #514, opened 2021-11-28
Login with Gitea OAuth problems: external user ID not properly recorded


gytisrepecka:
### Describe the bug
When Gitea OAuth is used for logging in, external user ID is not properly recorded in `oauth_users.remote_user_id` which causes misbehavior when logging in with different external users from Gitea server.

### Steps to reproduce
1. Login to WriteFreely with `user_a`;
2. Associate `user_a` with Gitea user `gitea_user_a`;
3. Login to WriteFreely with `user_b`;
4. Associate `user_b` with Gitea user `gitea_user_b`;
5. Attempt to login with Gitea user `gitea_user_b` - you will be logged in as `user_a` while it is expected to be `user_b`.

### Expected behavior
Expected behavior is logging in with `gitea_user_a` logs into `user_a` while `gitea_user_b` logs into `user_b`.

### Application configuration
 - **Single mode or Multi-user mode?** Multi-user
 - **Database?** [mysql/sqlite] MySQL
 - **Open registration?** [yes/no] Yes and No
 - **Federation enabled?** [yes/no] Yes

**Version or last commit**: v0.13.1

To solve this issue it is required to add extra auth fields into Gitea implementation code in the same way as they are available in `oauth_generic.go`. Pull request will be submitted in a moment.

gytisrepecka:
This is very related to #469.

gytisrepecka:
Pull Request fixes the problem, running for a while now in production. Can it be merged?

thebaer:
Fixed in #515

PR #515, opened 2021-11-28
#514 - Fix login with Gitea OAuth: external user ID not recorded

gytisrepecka:
Fix issue #514 - fix details:

* change inspect endpoint, add missing fields to receive scope, external user ID, username, display name and email;
* change error message when OAuth account is already associated;
* log external (Gitea) user ID in addition to internal (WriteFreely) user ID;
* include additional auth fields into `oauth_gitea.go` based on `oauth_generic.go`.

---

### DIFF
--- a/oauth.go
+++ b/oauth.go
@@ -293,10 +293,15 @@ func configureGiteaOauth(parentHandler *Handler, r *mux.Router, app *App) {
 			ClientID:         app.Config().GiteaOauth.ClientID,
 			ClientSecret:     app.Config().GiteaOauth.ClientSecret,
 			ExchangeLocation: app.Config().GiteaOauth.Host + "/login/oauth/access_token",
-			InspectLocation:  app.Config().GiteaOauth.Host + "/api/v1/user",
+			InspectLocation:  app.Config().GiteaOauth.Host + "/login/oauth/userinfo",
 			AuthLocation:     app.Config().GiteaOauth.Host + "/login/oauth/authorize",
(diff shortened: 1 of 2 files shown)


gytisrepecka:
Running successfully in production for a while now. Anyone could test it as well?

gytisrepecka:
> Thanks for addressing this, @gytisrepecka!
> 
> This looks good from reading through, but I have just one concern: in what Gitea version did the inspect endpoint change? I'm thinking this might break current integrations, and we'll probably want to document this somewhere so people are aware.

I believe I used wrong inspect endpoint from the beginning: in [Gitea OAuth2 provider documentation](https://docs.gitea.io/en-us/oauth2-provider/) **OpenID Connect UserInfo** is specified as `/login/oauth/userinfo`.

OAuth provider functionality is available in Gitea since v1.8.0 and it seems real endpoint did not change over time - it was simply [not specified](https://github.com/go-gitea/gitea/commit/29695cd6d511823ab0d233bba8c7971e5dac3e5f#diff-e92fdc50af1f3b627325df503decfef6204e717f30fee35bda9e625e50afcedf) in documentation, therefore I took [whatever I was able to find from API](https://gitea.com/api/swagger#/user/userGetCurrent).

thebaer:
Ah, okay great -- that makes sense!

thebaer (approved):
Since it's been working in production for a while, and code looks good, this is good to go :+1:

PR #520, opened 2021-12-21
Elide Markdown when generating slug

thebaer:
This makes sure e.g. link URLs and image alt text don't get included in the slug.

Closes [T329](https://phabricator.write.as/T329)

---


thebaer:
This is deployed and working well on Write.as. Merging now.

PR #521, opened 2021-12-25
Update 404 error message

eli-oat:
This is a small pull request to replace the current 404 error message. The current one implies that the error is the person visiting the site's fault, which is interpreted by some, especially in translation, as relatively hostile. This simplified, message makes it clear that the error is in software, and not the fault of a person.

---


eli-oat:
Bumping this

thebaer:
Sorry for the delay, and thanks for submitting this fix! Great point about this getting lost in translation. Merging now.

PR #528, opened 2022-01-12
Protect drafts if they are part of a Private or Protected collection

isaacsu:
- Draft posts remain publicly accessible even after they are moved to a Private or Protected collection. 
- This PR causes writefreely to return with a 404 when a non-owner tries to access a draft post that has been moved to a Private or Protected collection.

--- 


posts.go
```diff
@@ -424,6 +425,18 @@ func handleViewPost(app *App, w http.ResponseWriter, r *http.Request) error {
...
+	var protectDraft bool
+	if found && collectionID.Valid {
+		collection, err := app.db.GetCollectionByID(collectionID.Int64)
+		if err != nil {
+			log.Error("view post: %v", err)
+		}
+
+		protectDraft = (collection.IsPrivate() || collection.IsProtected())
+	} else {
+		protectDraft = true
```
thebaer, line 437: I believe this can be removed? Seems to make every single draft post private, even when it doesn't belong to a private / protected collection.

thebaer (changes requested):
Thanks for working on this, and sorry it's taken so long to review. Looks like some of the logic needs to be tweaked just a bit.

isaacsu:
Hi @thebaer, I've moved on from writefreely for the time being, so I won't be pursuing this PR any further. Happy for you to take over and run with it if you're keen. Regardless, thank you for creating writefreely and sharing so generously it with the world 🙏.

thebaer:
Thanks again for starting this @isaacsu! Merging this now.

thebaer (approved):

PR #537, opened 2022-03-03
Prevent dropping of external files onto editor

dariusk:
We now detect whether a file is being dropped by a user (by checking `event.dataTransfer.files.length`) and prevent the remaining drop behavior if this is the case. Otherwise, drop happens like normal (so a user can still drop text into the editor, or even an image that has been uploaded already via normal means and rendered in the editor).

This should address [this issue](https://discuss.write.as/t/posts-turn-into-scrambled-incomprehensible-text/4794/4).

---

### DIFF
--- a/prose/prose.js
+++ b/prose/prose.js
@@ -83,6 +83,14 @@ class ProseMirrorView {
         typingTimer = setTimeout(doneTyping, doneTypingInterval);
         this.updateState(newState);
       },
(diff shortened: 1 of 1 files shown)


thebaer (approved):
This works perfectly -- and good catch keeping the support for dropping in text. Thanks, @dariusk!

PR #543, opened 2022-04-27
Add unix socket support

### LINKED ISSUE #542: Listening on unix socket
Essentially, this will make it easier to connect WriteFreely up to reverse proxies like nginx, since the trip through the proxy doesn't have to use TCP.

I'm filing an issue here because I plan to submit a PR for this soon, since I don't think the amount of work required is very big. The main change would be to accept file paths for the listen port parameter, which would listen on that path as a unix socket instead of on a port via TCP.

clarfonthey:
Enables listening on unix sockets by specifying a file path for the bind address. Fixes #542.

### DIFF
--- a/app.go
+++ b/app.go
@@ -504,9 +504,8 @@ requests. We recommend supplying a valid host name.`)
 			err = http.ListenAndServeTLS(fmt.Sprintf("%s:443", bindAddress), app.cfg.Server.TLSCertPath, app.cfg.Server.TLSKeyPath, r)
 		}
 	} else {
-		var network string
-		var protocol string
-
+		network := "tcp"
+		protocol := "http"
 		if strings.HasPrefix(bindAddress, "/") {
 			network = "unix"
 			protocol = "http+unix"
@@ -519,8 +518,6 @@ requests. We recommend supplying a valid host name.`)
 				os.Exit(1)
 			}
 		} else {
-			network = "tcp"
-			protocol = "http"
 			bindAddress = fmt.Sprintf("%s:%d", bindAddress, app.cfg.Server.Port)
 		}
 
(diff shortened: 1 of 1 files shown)


thebaer:
Great addition -- thanks for submitting this! I'll test as soon as I can. In the meantime, anyone else should feel free to review!

clarfonthey:
I did run the automated tests and everything passed, although I haven't gotten around to testing the actual server yet. Technically other options for unix sockets could be added, but I figured I'd choose the option with the minimum surface area for now since I'm not sure how much that's desired.

Will add some extra remarks as needed once I get around to that.

clarfonthey:
Okay so, I finally got around to testing this, noticed a few issues, and fixed them. Notable changes:

1. Error handling in golang is awful and I accidentally shadowed `err`, which is now fixed
2. The server needs to delete any leftover socket files in order to actually listen on the socket; if it doesn't have permission to do this, it will just throw an error
3. I found out that `http.Serve(listener, r)` exists so I did that instead of `(&http.Server{Handler: r}).Serve(listener)`

thebaer:
I went ahead and cleaned the code up to be a little more idiomatic, and made it so the socket file is deleted on server shutdown.

I probably undid your error-naming changes -- I'm not sure what the error shadowing issue was that you mentioned, since past code was overwritten by force-push. But let me know if this causes any issues! Otherwise I tested and it works well, so I'll merge this soon.

clarfonthey:
I don't remember what the error change was, but I was particularly frustrated with the compiler in that moment and the brute force approach worked. So, if your code works, that's good.

I didn't remove the socket initially since I figured it didn't matter much since it'd be overwritten on the next time it starts, but it is an improvement. Thank you.

Everything looks good to me. Appreciate you tidying this up for me.

thebaer:
Merging this now. Thanks again for working on this!

PR #553, opened 2022-05-25
Create container from GitHub action

### LINKED ISSUE #518: docker tags are not up to date
Hey, 
I wanted to run writefreely as a docker container but noticed that there are only the `latest` tags and the `0.12.x` tags. 
However the latest writefreely version is `0.13.1`. 

Would be nice if you could fix that and add a `0.13.1` tag. 
(Note that the `latest` tag does not suffice here, as it may point to `0.14.x` or even `1.x.x` in the future, which might break deployments due to breaking changes in writefreely) 

As a suggestion it would also be cool if you introduced `0.x` tags, so that one could automate upgrading to never hotfix versions.

### LINKED ISSUE #547: Outdated docker-images posted on docker hub
### Describe the bug
The container-image on docker hub is 10 months old, and is pressumably severely out of date with regards to support libraries and general patch-management for the base-container-image.

Since this is a web-facing application, that should probably be remediated.

I propose a thing where the container-tag `latest` points to the latest development in the `develop`-branch, and release-tags are added as tagged here on github.

I wanted to propose this with a PR, but I couldn't find the relevant parts for how the images are put up on docker hub, can I assume that this process is manual, and/or depends on the, no longer free, auto-build on docker hub?

I'm prepared to fix some parts of this myself, as from other issues here, I see you claim to not be as familiar with container images - I need to know how to integrate with the current processes however.

davralin:
As mentioned in #547  - this will create a GitHub-action that will create container-images automatically on:
  - Merge to develop - which should end up with the :develop-tag.
  - Tagged release - which should end up with the  :<tag>-tag. (i.e. :v0.13.2)
 
The containers will be available on GitHub's repository.

In addition, this PR:
 - updates the running image to be the latest alpine:3-image, instead of a hardcoded older version.
 - adds dependabot-alerts for the Dockerfile, and github-actions - to ensure future updates.

This PR does NOT:
  - Update the container hosted on Docker Hub.
    That needs credentials set as a secret in this repo, and I haven't had the time to check which is needed.

Would like to get this underway first, so people can start testing the image, and then I'll get a new PR going for updating docker - I see that I need to do some updates in the documentation-repo also.

Closes #547
Closes #518 

---

- [X] I have signed the [CLA](https://phabricator.write.as/L1) (But I didn't get an email)


.github/workflows/docker-publish.yml
```diff
@@ -0,0 +1,61 @@
+name: Build container image, publish as Github-package
+
+# This workflow uses actions that are not certified by GitHub.
+# They are provided by a third-party and are governed by
+# separate terms of service, privacy policy, and support
+# documentation.
+
+on:
+  push:
+    branches: [ master, develop ]
```
thebaer, line 10: Our main release branch is `main`, instead of `master` -- think we'll want to change this?

thebaer (changes requested):
Thanks for submitting this! Looks good from my understanding -- just one change I think we need to make.

.github/workflows/docker-publish.yml
davralin: Right, I knew that.

PR #585, opened 2022-10-02
Truncate Post created times down to 1 second

moredhel:
This fixes an issue with the writefreely ios application.

Swift is unable to parse date-times with fractions of a second included.

This has 2 changes.

1. On all now posts created, we truncate time-created down to the second
2. A migration script to update the 'created' column to strip out the milliseconds.

---


moredhel:
This issue has been resolved :)

PR #613, opened 2022-12-26
Replace mattn/go-sqlite3 with modernc.org/sqlite

lpar:
This replaces the CGO wrapped C SQLite from [github.com/mattn/go-sqlite3](https://github.com/mattn/go-sqlite3) with the pure Go implementation of SQLite from [modernc.org/sqlite](https://pkg.go.dev/modernc.org/sqlite). This should remove the need for CGO to build WriteFreely. That is you can do:

```
CGO_ENABLED=0 go build -v -tags='sqlite' ./cmd/writefreely/
```

I've tested that I can initialize a new database, and also that I can load and serve the web UI from a database that was created with a build of WriteFreely using the CGO SQLite. Also, the tag pages work, which suggests that the add-in regexp command works. However, obviously this could all use some more extensive testing by people with significant SQLite databases.

Motivation: With CGO SQLite removed, it should be possible to cross-compile WriteFreely without needing `xgo` or a C compiler setup. Not all OS and architecture combinations are supported by modernc.org/sqlite, but [the list includes all the major ones](https://pkg.go.dev/modernc.org/sqlite#hdr-Supported_platforms_and_architectures). So this should make it easier to compile on a Mac and deploy on Linux, build binaries for all platforms using CI/CD, and so on.

---


thebaer:
Thanks for submitting this! It would be great to move to this in the future.

For now, the only blocker for me is that it doesn't support `arm` architectures the Raspberry Pi uses -- and that's kind of a selling point for the software. So I think we'll hold off on merging this until that support is there in modernc.org/sqlite.

lpar:
According to https://pkg.go.dev/modernc.org/sqlite the package supports `arm` for Linux, and `arm64` for Linux, BSD, Darwin and Windows. I think that's everything you're likely to encounter on a Raspberry Pi. RetroPi is `arm` (or at least, mine is); Raspbian / Raspberry Pi OS now supports `arm64`. Darwin and Windows only support `arm64` anyway.

I'm not aware of any big-endian ARM Linux distributions for the Pi, threads I can find suggest that it's not trivial to get Linux to work in big-endian mode because the hardware starts up in LE and has to be flipped to BE.

PR #618, opened 2023-01-03
Widen oauth_users.access_token to prevent insertion failure

Timshel:
Hi,

I configured an instance using mariadb and [Keycloak](https://www.keycloak.org/) for SSO.
User creation was failing since the returned access token is longer than the current field in db (`VARCHAR(512)`)

I believe switching to `TEXT` is better than setting an arbitrary length, overhead should be minimal especially since I don't believe it's in any index.

I added a conditional to  run it only on `MySQL` since I believe `SqlLite` only use `TEXT`.
    
---


thebaer:
Thanks for submitting this! And sorry for the delay. Looks good to me -- I'll merge this and get it in the next release.

PR #619, opened 2023-01-03
Additionnal config for oauth registration

Timshel:
Hi,

I configured an instance using [Keycloak](https://www.keycloak.org/) for SSO but I wanted to be able to allow registration after a successful authentication without opening it to everyone or using an invitation code.

When using an external oauth provider I believe it makes sense to be able to control registration independently of the global setting.

Added `allow_registration` to external oauth configs to allow the ability to control this logic.
I omitted it from the `WriteAsOauth` config since I believe it does not make sense, just reused `OpenRegistration` when I needed to set a parameter in  `configureWriteAsOauth`. 
I added the `AllowRegistration` control to the `oauthHandler`  since it makes the change way less verbose than adding an additional function to all the different `oauthClient`.

---

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -76,47 +76,51 @@ type (
 	}
 
 	GitlabOauthCfg struct {
-		ClientID         string `ini:"client_id"`
-		ClientSecret     string `ini:"client_secret"`
-		Host             string `ini:"host"`
-		DisplayName      string `ini:"display_name"`
-		CallbackProxy    string `ini:"callback_proxy"`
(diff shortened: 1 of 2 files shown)


natsukagami:
This is very useful for me.

natsukagami:
@Timshel it seems you can do exactly this by setting
```ini
disable_password_auth = true
open_registration = true
```

Timshel:
@natsukagami Oh might have missed the `disable_password_auth` conf will have to test it thx :)

Issue #85, opened 2019-03-24
WriteFreely runs as root


gytisrepecka:
Based on system service configuration file provided in [production documentation](https://writefreely.org/start#production), **writefreely** binary will run under **root** system user. To fix this, make sure to add following parameters in **Service** section:
`User=writefreely`
`Group=writefreely`
After systemctl daemon is reloaded, you will see writefreely binary now runs under corresponding user:
`ps aux | grep writefreely`

gytisrepecka:
Full configuration file here: https://gist.github.com/gytisrepecka/557660aed94afebec68c0a2432ae348e

thebaer:
For this, we'll also need to update the documentation to include the creation of the `writefreely` user, right? Would you mind creating an issue for this in the [documentation repo](https://github.com/writefreely/documentation)?

gytisrepecka:
Yes, I can document that.

thebaer:
Going to close this out since it isn't a bug. Let's continue discussing this on the documentation repo.

cburkert:
Its 2023 and the installation guide still leads users to an instance running fully as root.
I couldn't find any continuation of this discussion on the documentation repo.
Could anyone please give me a pointer or did the discussion end here?

PR #630, opened 2023-01-23
Abort startup if run as root

cburkert:
## The Problem
Running services permanently as `root` is considered a bad practice, because any vulnerability in the service immediately poses the threat of full system takeover.

The [installation instructions](https://writefreely.org/start) for Writefreely tell users to setup services (e.g., via systemd) that run as root. This fact has been reported and criticised here before (#85), but unfortunately without any consequences.

## The Solution
This patch prevents Writefreely from being run as root. It simply aborts with an error if you try to do so.

## Notes
To accompany this patch, the installation instructions should be changed to use non-root service UIDs/GIDs.
I'd be happy to provide pull requests for that as well, but this part of the documentation seems out of scope of the documentation repo.

For standalone operation of writefreely, the best practice to allocate privileged ports like (80/443) would be to use `setcap` on Linux systems. Alternatively, writefreely could consider dropping privileges after their allocation.

---

### DIFF
--- a/cmd/writefreely/main.go
+++ b/cmd/writefreely/main.go
@@ -14,6 +14,7 @@ import (
 	"fmt"
 	"os"
 	"strings"
+	"syscall"
 
 	"github.com/gorilla/mux"
 	"github.com/urfave/cli/v2"
@@ -22,6 +23,10 @@ import (
 )
 
 func main() {
+	if syscall.Getuid() == 0 {
+		log.Error("Running as root is insecure. Use setcap for privileged resource access")
+		os.Exit(1)
+	}
 	cli.VersionPrinter = func(c *cli.Context) {
 		fmt.Printf("%s\n", c.App.Version)
 	}


thebaer:
Thanks for submitting this. My main concern is that this will aggravate users who really don't need this level of protection, e.g. when just testing the software out, running it on a local machine or private network, etc. Forcing _everyone_ to know the ins and outs of `setcap` when they just want to get this up and running I think will make them give up and stop trying to install the software at all. I'm also not aware of any other web server software that does this to users (please let me know if I'm wrong).

In my opinion, this would be fine as just a strong warning message that gets logged when the server starts up. I think it's primarily a documentation issue that we should address in the install guide.

But if we're going to actively prevent people from running the software, we need to provide an override flag. And when the program fails to start, we need to instruct users what that flag is, so they can override these protections. ...And considering that, thinking this experience through as an admin and how annoying that would be, again I think we could solve this by just updating the docs and encouraging best practices there.

cburkert:
I get your concern about this potentially aggravating users. But, honestly, I cannot think of any sensible reason why anyone should run Writefreely as root. Hence aggravating, or rather protecting, users is kind of the point.
The general idea behind security by design is that doing the safe thing should be easy and default, while doing the dangerous thing should be hard and deliberate.

To your question how other services handle this issue: Traditionally, daemons and services used the approach of _privilege de-escalation_ to follow the paradime of _least privilege_. Software like Apache or nginx are started as root, allocate privileged resources and drop-down to non-privileged UIDs (e.g., www-data). Often there are controller processes that remain privileged and worker processes that get de-escalated.
This type of in-process de-escalation seems to be less practical with for software written in languages that have runtimes like Go. The webserver Caddy had also [a lengthy discussion](https://github.com/caddyserver/caddy/issues/528) about this and finally desided to do nothing.

I mean one could argue that a secure setup is up to the operator of the software and that they should be at liberty to run Writefreely as root if they want to.
But I would argue in the direction of security by default: Since Writefreely does not de-escalate privilege as many admin might expect from older (more traditional) software, it has a responsability to protect their users against unwanted risks.
Especially considering the potentially large install-base that is running as root without knowing, making the service abort by default would be a good way force them into changing their insecure installation.
Introducing a special flag to overwrite the abort would be a good compromise, I guess.

thebaer:
> But, honestly, I cannot think of any sensible reason why anyone should run Writefreely as root.

I mentioned a few reasons why they might above.

I get your perspective from a pure-security standpoint. But I need to consider all aspects of this product. And to me, a 100% secure product that no one can start up is worse than one that's less secure-by-default, but people can easily start up (and still easily make secure).

Less philosophically, this change will probably break some existing installations, so we can't introduce it without caring for them. Otherwise, we'll be taking down entire WriteFreely communities, forcing admins to spend an entire day on upgrading, or blocking them from upgrading at all -- which again accomplishes nothing, including better security.

Instead of merging this PR, I'd suggest we update the documentation first -- and I'd appreciate your help with that if you're willing. Then we can see if / what changes to the actual application are necessary.

cburkert:
Make no mistake, I don't want to downplay the impact this might have on your installed base. In the end, its a trade-off that you have to make, whether you take the risk of a potentially very severe vulnerability impact or the aggravation of the admins. Not an easy decision to make.

But I think you exaggerate the added barrier of entry that a mandatory non-root setup would have for new or existing users. Setting the necessary capabilities on the executable or in a systemd config is just one additional step in the setup process.
Even users that just want to test the software should be motivated enough to do so.

But if you decide not to prevent running as root, then I very much would second your idea to log a very strong warning message, which links to the documentation on how to securely setup the service.
Regarding the documentation, I am willing to help and will open an issue there shortly.

gytisrepecka:
Since Go code can be built into a binary on most operating systems, we should not have OS-specific logic within application itself. Will `syscall.Getuid()` run on and return expected value on, say, Windows build?
I totally agree `writefreely` should always run on unprivileged user, but that is up to system administrator to enforce.

PR #638, opened 2023-01-31
Strip debugging information

zer-far:
Using `go build -ldflags="-s -w"` reduces the file size of the binary by about 11 MB (from 39 MB to 28 MB).

---


thebaer:
Thanks for submitting this.

I'm not familiar with these build options -- is this going to make crash reports less useful? I think ideally we'd keep whatever info we need in the executable to get full stacktraces whenever the app panics, to be sure we can debug issues that come up.

zer-far:
I'm not sure but according to [https://pkg.go.dev/cmd/link](https://pkg.go.dev/cmd/link), `-s` would "Omit the symbol table and debug information." and `-w` would "Omit the DWARF symbol table.".

thebaer:
Got it. Before I can merge this, someone will need to confirm that when the application `panic()`s it still outputs useful information.

thebaer:
Thanks again for this! Merging now.

PR #639, opened 2023-02-01
Update modules

zer-far:
I have used the [go-mod-upgrade](https://github.com/oligot/go-mod-upgrade) tool to update go.mod and go.sum.

---


thebaer:
Thanks, but dependencies are upgraded individually, with automated PRs like #641.

If you're interested in helping with this, it would be great to have someone review the changes in those PRs and comment on whether the upgrade seems minor (and probably doesn't break anything), or if there might be breaking changes with any new versions.

PR #660, opened 2023-03-14
Loosen restrictions on rendered img alt attribute

### LINKED ISSUE #649: Several characters in the alt text result in no alt text at all
### Describe the bug
Having certain characters in the alt text (as far as I tested, these are `:`, `"`, `'`) makes the alt-text on images not render. I suppose it's a parsing error returning `undefined` or an empty string.

1. Mak a new post
2. Add an image `[My alt text. That's very "funny".](/path/to/image)`
3. Open the rendered post
4. The alt text ist missing
5. Add another image `[My alt text. That is very funny.](/path/to/image)`
6. The alt text is there

### Expected behavior
The alt text should render regardless of the given input as long as it is compliant with W3 standards. 
https://html.spec.whatwg.org/multipage/embedded-content.html#dom-img-alt

### Application configuration
 - **Single mode** 
 - **Database?** [write.as]
 - **Open registration?** [yes]
 - **Federation enabled?** [yes]

**Version or last commit**:

thebaer:
Previously, certain characters weren't allowed, and they would cause the entire alt attribute to be elided from the rendered page. Since we safely sanitize the content of this attribute anyway, this is unnecessary, so we now allow all text entered there.

Fixes #649

---

### DIFF
--- a/postrender.go
+++ b/postrender.go
@@ -270,6 +270,7 @@ func getSanitizationPolicy() *bluemonday.Policy {
 	policy.AllowAttrs("target").OnElements("a")
 	policy.AllowAttrs("title").OnElements("abbr")
 	policy.AllowAttrs("style", "class", "id").Globally()
+	policy.AllowAttrs("alt").OnElements("img")
 	policy.AllowElements("header", "footer")
 	policy.AllowURLSchemes("http", "https", "mailto", "xmpp")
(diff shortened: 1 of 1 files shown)


thebaer:
This is in production on Write.as and working well. Merging now.

PR #694, opened 2023-05-11
Fix compatibility issues with MySQL 8.0.4+ regex

fnyu:
As per MySQL documentation (https://dev.mysql.com/doc/refman/8.0/en/regexp.html):

>MySQL implements regular expression support using International Components for Unicode (ICU), which provides full Unicode support and is multibyte safe. (Prior to MySQL 8.0.4, MySQL used Henry Spencer's implementation of regular expressions, which operates in byte-wise fashion and is not multibyte safe. For information about ways in which applications that use regular expressions may be affected by the implementation change, see [Regular Expression Compatibility Considerations](https://dev.mysql.com/doc/refman/8.0/en/regexp.html#regexp-compatibility).

In the same document it states:

>The Spencer library supports word-beginning and word-end boundary markers ([[:<:]] and [[:>:]] notation). ICU does not. For ICU, you can use \b to match word boundaries; double the backslash because MySQL interprets it as the escape character within strings.

Fixes: https://github.com/writefreely/writefreely/issues/615
Fixes: https://github.com/writefreely/writefreely/issues/614

---

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -63,6 +63,7 @@ type (
 		Host     string `ini:"host"`
 		Port     int    `ini:"port"`
 		TLS      bool   `ini:"tls"`
+		IcuRegex bool   `ini:"icu_regex"`
 	}
 
 	WriteAsOauthCfg struct {
--- a/config/setup.go
+++ b/config/setup.go
@@ -12,12 +12,13 @@ package config
 
 import (
 	"fmt"
+	"strconv"
+	"strings"
+
 	"github.com/fatih/color"
 	"github.com/manifoldco/promptui"
 	"github.com/mitchellh/go-wordwrap"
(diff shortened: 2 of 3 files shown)


thebaer:
Thank you for fixing this and documenting it! It would be great to resolve this.

Do you think there's a way we could solve this automatically, without adding a new configuration value? My concerns are:

* Configuration value bloat
* I don't know what this value actually means without reading docs or knowing the intricacies of MySQL regex
* Things will unexpectedly break if an admin upgrades their MySQL db without also manually changing this value
* Admins upgrading WF will also need to manually change their config to fix this issue -- ideally, it'd be automatic
* We also need to consider Mariadb
* With the current MySQL version, this should probably default to `true`? But of course might be backwards compatibility issues

PR #730, opened 2023-07-11
chore: slice replace loop

testwill:
[activitypub.go:647:3:  replace loop with na.CC = append(na.CC, instFolls...)
activitypub.go:716:3:  replace loop with na.CC = append(na.CC, instFolls...) 
db/create.go:250:2:  replace loop with things = append(things, b.Constraints...)]

---


thebaer:
Thanks again, merging now!

PR #734, opened 2023-07-29
Add file uploading and file sharing functionality

alborzjafari:
- I need to upload media files such as images and use them in markdown format. I have added this feature in this pull request.  
- It is configurable using config.ini file to activate or deactivate this feature in the instance. **It is disabled by default please enable it for testing in config.ini (`allow_upload_media = true`)**
-  Size of the files can be limited using `media_max_size`, please set `media_max_size` in config.ini before running. Its unit is megabyte, for example 10 megabytes:
```
media_max_size = 10 
```
- The uploaded media can be shared using markdown for example:
```
![picture](/media/<author>/<post_name>/<uuid>.png
or
![ducument](/media/<author>/<post_name>/<uuid>.pdf
```
- Uploading media files is possible in _Edit Post Metadata_ section. (edit the post-> edit post metadata or here: http://<example.com>/<author_name>/<post_name>/edit/meta), also the list of uploaded files is there.
- The front-end part needs more improvements.
- Total space can be limited using `total_media_space`, its unit is megabyte(s).
---

### DIFF
--- a/pad.go
+++ b/pad.go
@@ -337,6 +337,24 @@ func handleGetFile(app *App, w http.ResponseWriter, r *http.Request) error {
 	return nil
 }
 
+func calculateDirectoryTotalSize(dirPath string) (int64, error) {
+	var totalSize int64
+
+	err := filepath.Walk(dirPath, func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+
+		if !info.IsDir() {
+			totalSize += info.Size()
+		}
+
+		return nil
+	})
+
+	return totalSize, err
+}
+
 func handleUploadMedia(app *App, w http.ResponseWriter, r *http.Request) error {
 	maxUploadSize := app.cfg.App.MediaMaxSize * 1024 * 1024
 	r.Body = http.MaxBytesReader(w, r.Body, maxUploadSize)
@@ -346,6 +364,8 @@ func handleUploadMedia(app *App, w http.ResponseWriter, r *http.Request) error {
 		http.Error(w, errMsg, http.StatusBadRequest)
 		return nil
 	}
+
+	fileSize := r.ContentLength
 
 	if err := okToEdit(app, w, r); err != nil {
 		 return err
@@ -372,8 +392,17 @@ func handleUploadMedia(app *App, w http.ResponseWriter, r *http.Request) error {
 	err = os.MkdirAll(mediaDirectoryPath, 0755)
 	if err != nil {
 		return err
-	} else {
 	}
+
+	totalSize, err := calculateDirectoryTotalSize(mediaDirectoryPath)
+	totalMediaSpace := app.cfg.App.TotalMediaSpace * 1024 * 1024
+	if totalSize + fileSize > totalMediaSpace {
+		errMsg := fmt.Sprintf("Your upload space limit has been exceeded. Your limit is: %d MB",
+					app.cfg.App.TotalMediaSpace)
+		http.Error(w, errMsg, http.StatusBadRequest)
+		return nil
+	}
+
 	newFileName, _ := getNewFileName(mediaDirectoryPath, handler.Filename)
 	newFilePath := filepath.Join(mediaDirectoryPath, newFileName)
 	dst, err := os.Create(newFilePath)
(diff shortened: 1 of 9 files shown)


sssilver:
This is such a huge feature. Can we expedite reviewing this and getting this merged?

alborzjafari:
Could you please review this pull request?

thebaer:
Thanks for submitting this. This is a pretty huge feature with many repercussions for the product, so this will need to be pretty solid before we can merge. I'll leave comments in the code as well, but here are some important things that need to be addressed:

* Media metadata should probably be stored in the database, to help with management and permissions. Note that we've done a ton of work on this with our Write.as instance and Snap.as ([using this table schema](https://paste.as/0d0efk3wnkifir4j)), and we could build on this if you want to collaborate on this instead of writing from scratch.
* We'll need accompanying documentation in https://github.com/writefreely/documentation to ensure people configure any reverse proxy to also allow large file uploads (e.g. most servers are configured for <10MB by default)
* Please run `go  fmt` on all files, per the [Code Guidelines](https://github.com/writefreely/writefreely/blob/develop/CONTRIBUTING.md#code-guidelines)

Again if you'd like to work on this together, another way forward would be for you to work only on the UI for this -- uploading images and managing them -- and then I could contribute the backend work that we've already put in. Either way, let me know what you'd like to do.

pad.go
```diff
@@ -185,9 +277,163 @@ func handleViewMeta(app *App, w http.ResponseWriter, r *http.Request) error {
...
+	}
+	file, handler, err := r.FormFile("file")
+	if err != nil {
+		http.Error(w, "Error retrieving the file", http.StatusInternalServerError)
+		return nil
+	}
+	defer file.Close()
+	user := getUserSession(app, r)
+	mediaDirectoryPath := filepath.Join(app.cfg.Server.MediaParentDir, mediaDir,
+					user.Username, slug)
```
thebaer, line 404: Storing media in a directory with the user's username will break when they change their username. This should use something stable, like their user ID.

pad.go
```diff
@@ -185,9 +277,163 @@ func handleViewMeta(app *App, w http.ResponseWriter, r *http.Request) error {
...
+		return nil
+	}
+	defer file.Close()
+	fileInfo, err := file.Stat()
+	if err != nil {
+		http.Error(w, "Failed to get file information", http.StatusInternalServerError)
+		return nil
+	}
+	w.Header().Set("Content-Disposition", "attachment; filename="+fileInfo.Name())
+	w.Header().Set("Content-Type", "application/octet-stream")
```
thebaer, line 339: Files should probably be delivered as their correct MIME type, instead of `application/octet-stream`. There are default Go packages that help with this.

alborzjafari:
1. "Media metadata should probably be stored in the database, to help with management and permissions." : I tried to to do something like inheriting permissions of a post for its files.
2. "...another way forward would be for you to work only on the UI for this -- uploading images and managing them...": Would you address me in source codes, I didn't see about uploading images. My purpose is uploading all kind of files not only the images.

pad.go
```diff
@@ -185,9 +277,163 @@ func handleViewMeta(app *App, w http.ResponseWriter, r *http.Request) error {
...
+	if err := r.ParseMultipartForm(maxUploadSize); err != nil {
+		errMsg := fmt.Sprintf("File size limit exceeded. The limit is: %d MB",
+					app.cfg.App.MediaMaxSize)
+		http.Error(w, errMsg, http.StatusBadRequest)
+		return nil
+	}
+
+	fileSize := r.ContentLength
+
+	if err := okToEdit(app, w, r); err != nil {
```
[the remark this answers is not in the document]
alborzjafari: In some cases, it is necessary to check the situation before uploading a file, especially for silenced users. While the UI may not display any issues, using Postman or similar tools can allow file uploads for any post by a silenced user. The purpose of using this function is to prevent such problems.

antranigv:
Any news on this? My users keep bugging me that they want this :)

Cheers, y'all!

pwd-github:
> Any news on this? My users keep bugging me that they want this :)
> 
> Cheers, y'all!

I too would like to get a status update about this

thebaer:
We're going to support uploads in WriteFreely, both [generic files](https://todo.musing.studio/T929) and [images](https://todo.musing.studio/T688). It's all currently planned for our v0.17 release. However, this PR doesn't go in the direction we need to support that, so at this point, I'm going to close it. Instead, we'll use the extensive functionality we've developed for Write.as, Snap.as, etc. and bring that into WF in that upcoming version.

Issue #612, opened 2022-12-26
Editor Screen Doesn't Support Bidirectional Texts


UsernamesAreFuckedUp:
### Describe the bug
When I'm writing in an RTL language at the editor screen, text direction is not set to RTL. But it has no problems with LTR languages. In other words, the user is expecting the text field to support bidirectional texts, but it doesn't. 

### Steps to reproduce (if necessary)
1. Go to the editor screen
2. Start typing in an RTL language (Persian, Arabic, etc.), followed by some LTR Paragraph.
3. You'll see something like this, which is disturbing.
![image](https://user-images.githubusercontent.com/96365880/209567446-fe6b0612-657f-4a63-9195-987db399cbe2.png)

### Expected behavior
![image](https://user-images.githubusercontent.com/96365880/209567203-f888bcc3-19e7-406f-849e-7c0a98615070.png)
Adding bidi support is very easy, I assume. I just attached `dir="auto"` to the `textarea`. (Maybe there are some other places you should use this attribute. I don't know.)

### Applications configuration
 - Multi-user Mode
 - Open Registration
 - Federation Enabled

v0.13.1

thebaer:
Thanks, that fix should be easy enough! Anyone should feel free to tackle this. Otherwise I'll get to it when I can.

PR #735, opened 2023-08-01
Adjust alignment for editing RTL pages

alborzjafari:
Editing RTL pages is not appropriate and makes it difficult to edit an RTL page, especially when using a LTR language with RTL. This pull request aims to address this issue.

---

### DIFF
--- a/templates/pad.tmpl
+++ b/templates/pad.tmpl
@@ -14,7 +14,7 @@
 
 		<div id="overlay"></div>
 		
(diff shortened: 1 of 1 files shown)


thebaer:
Thanks for this submission, but it seems there was a broader issue with the editors that occurred when composing new posts as well -- not just editing (see #612). That was fixed in #826, so I'll close this PR.

PR #764, opened 2023-09-17
Add South Azerbaijani language.

alborzjafari:
This pull request adds [South Azerbaijani](https://iso639-3.sil.org/code/azb) language.

---


thebaer:
Thanks for looking into this! However, the software only supports storing 2-letter language codes (this is the max length for the `language` column in the database), and we already have an `az` option. Is that sufficient, or do we need to look at supporting `azb` as well?

alborzjafari:
I need `azb` support, because they are different in writing, `az` is Latin and left to right but `azb` uses Arabic characters and is right to left.

PR #766, opened 2023-09-21
Correctly respond to application/ld+json requests

### LINKED ISSUE #564: Serving ActivityPub objects
### Describe the bug

According to [ActivityPub specification](https://www.w3.org/TR/activitypub/#retrieving-objects), servers should respond with ActivityStreams object representation to requests having `application/ld+json; profile="https://www.w3.org/ns/activitystreams"` in Accept header, but apparently only `application/activity+json` is supported

>Servers ... MUST present the ActivityStreams object representation in response to application/ld+json; profile="https://www.w3.org/ns/activitystreams", and SHOULD also present the ActivityStreams representation in response to application/activity+json as well.

https://github.com/writefreely/writefreely/blob/b7f732b9154096d6864f2fa9469fe9c7a9b73d58/posts.go#L1123

```shell
curl -H 'Accept: application/ld+json; profile="https://www.w3.org/ns/activitystreams"' https://***/api/posts/***
```

The server returns the standard API response:

```json
{
  "code": 200,
  "data": {
    ...
  }
}
```

### Expected behavior

Server should return ActivityStreams object representation.

thebaer:
This returns ActivityStreams objects when the Accept header is `application/ld+json; profile="https://www.w3.org/ns/activitystreams"`, per the ActivityPub spec.

Fixes #564

---

### DIFF
--- a/collections.go
+++ b/collections.go
@@ -496,8 +496,7 @@ func apiCheckCollectionPermissions(app *App, r *http.Request, c *Collection) (in
 
 // fetchCollection handles the API endpoint for retrieving collection data.
 func fetchCollection(app *App, w http.ResponseWriter, r *http.Request) error {
-	accept := r.Header.Get("Accept")
-	if strings.Contains(accept, "application/activity+json") {
+	if IsActivityPubRequest(r) {
 		return handleFetchCollectionActivities(app, w, r)
 	}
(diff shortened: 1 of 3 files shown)


thebaer:
Merging now!

PR #769, opened 2023-09-21
Fix monthly active user stats

thebaer:
Previously, an error in the query prevented the stat from ever showing up. This fixes that and adds logging.

---


thebaer:
Merging now

PR #771, opened 2023-09-21
Support rel=me verification on blogs

thebaer:
This adds a blog customization option that lets users set a URL or fediverse handle as a profile to verify. Then when linking to the blog from e.g. Mastodon, it will show as verified there. It does this by adding a `<link>` element to the page's `<head>` -- i.e. it doesn't show up anywhere on the blog itself.

To use this, you'll need to run `writefreely db migrate`.

Closes [T744](https://phabricator.write.as/T744)

---


thebaer:
Tested on a few instances and seems to be good. Merging now

PR #774, opened 2023-09-25
Fix fonts not getting applied on first load

### LINKED ISSUE #313: Fonts aren't applied when the network speed is low
### Describe the bug
When displaying the site on a device with a low network speed, the custom fonts are loaded afterwards and aren't applied.

Use the dev tools of your browser to throttle the connection quality to 3G-like.

See [this screencast](https://cloud.augendre.info/s/pBxG5k6xdSmc4Sw).

### Expected behavior
Either fonts should be applied after being downloaded or they should not be downloaded at all. Right now it's just a waste of resource.

### Application configuration
 - **Single mode or Multi-user mode?** Single
 - **Database?** [mysql/sqlite] sqlite

**Version or last commit**: 037fc40

thebaer:
In many cases, such as slow connections or disabled caching, custom fonts don't get applied, making the page look worse.

Fixes #313

---

### DIFF
--- a/less/fonts.less
+++ b/less/fonts.less
@@ -3,7 +3,6 @@
 	font-family: 'Open Sans';
 	font-style: normal;
 	font-weight: 400;
-	font-display: optional;
 	src: url('/fonts/open-sans-v13-latin-regular.eot'); /* IE9 Compat Modes */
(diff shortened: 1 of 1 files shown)


thebaer:
Merging now

PR #777, opened 2023-09-25
Support resetting password via email

thebaer:
This adds a self-serve password reset page. Users can enter their username and receive an email with a link that will let them create a new password. Once they've set a new password, they'll be directed to log in normally.

If they've never set a password, it will send them a one-time login link (building on #776) that will then take them to their Account Settings page. If they don't have an email associated with their account, they'll be instructed to contact the admin, so they can manually reset the password.

Includes changes to the stylesheet and database, so run:

    make ui
    writefreely db migrate

Closes [T508](https://writefreely.org/tasks/508)

_Remaining work:_

- [x] Disable the form if email isn't configured
- [x] Display a friendly message on `/reset` if email isn't configured
- [x] Display a friendlier error on form submission if email isn't configured
- [x] Consider a configuration with no passwords allowed, and disable, etc.

---

### DIFF
--- a/account.go
+++ b/account.go
@@ -1321,6 +1321,11 @@ func doAutomatedPasswordChange(app *App, userID int64, newPass string) error {
 func handleResetPasswordInit(app *App, w http.ResponseWriter, r *http.Request) error {
 	returnLoc := impart.HTTPError{http.StatusFound, "/reset"}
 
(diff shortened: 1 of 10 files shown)


thebaer:
Merging now.

Issue #783, opened 2023-10-06
Server error when accessing tags page


blujan:
### Describe the bug
I'm running Writefreely in a container with sqlite as DB, and I've come across what I think is a minor bug in the database API. When trying to access the page for tags (\<site\>/tag:\<tag\>), the server error page comes up instead of the list of posts with the tag as it should. Dropping into the logs I see the following error - 

```
2023-10-05 22:08:32 ERROR: 2023/10/06 05:08:32 database.go:1260: Failed selecting tagged posts: no such function: NOW
```

### Steps to reproduce (if necessary)
1. Setup a new instance using sqlite as backend
2. Create and publish a post containing a tag
3. Log out
4. Click on the tag link from within the post, i.e., \<uri\>/tag:\<tag\>

### Expected behavior
What should've happened? Page showing posts containing the given tag

### Application configuration
 - **Single mode** 
 - **Database?** sqlite
 - **Open registration?** no
 - **Federation enabled?** no

**Version or last commit**: 0.14.0

blujan:
So dipping into the code, I do see the following at database.go:GetAllPostsTaggedIDs() - 

```
	if !includeFuture {
		timeCondition = "AND created <= NOW()"
	}
```

Should it be calling the helper function db.now() instead since I don't see the sqlite check after this bit modifying this string? If I get some time, I'll try this out to see if it changes the behavior of the above issue.

thebaer:
Thanks for reporting this! Yes it should be calling `db.now()` here, instead. We'll gladly accept a PR if you do get around to fixing this!

PR #785, opened 2023-10-07
Fix use of NOW() when getting tagged posts

blujan:
Quick fix to address the issue in #783 - changes the query in the function that selects tagged posts to use the helper `db.now()` to correctly format for either SQL or sqlite.

I did A/B testing (exact same config, only different binaries) of the build with the fix compared the 0.14.0 release to confirm that this completely addresses the issue.

---


thebaer (approved):
Thanks for submitting a fix! This works great. Merging now.

PR #799, opened 2023-10-25
Conditionally use twitter large summary card format when an image is available

heyakyra:


thebaer (approved):
This works great, thanks @heyakyra!

PR #828, opened 2023-11-18
Dockerfile

d4rklynk:
Changes:

- Slightly deacreased the size of the docker image by removing some layers
- Added HEALTHCHECK
- Updated Golang and Alpine images to latest versions
- Changed FROM's value to sha256 for better accuracy
- Added `apk upgrade` to both images
- Added quotes in LABEL's value

---


thebaer:
Thanks for contributing this! Merging now.

PR #854, opened 2024-01-10
Fix Collection property serialization on API

thebaer:
Use standard string instead of sql.NullString for `style_sheet`, `script`, and `signature`.

Addresses #820

---


thebaer:
Merging now!

PR #874, opened 2024-02-06
Version number and healthcheck fixes for Docker image

### LINKED ISSUE #873: Version not available in Docker image
### Describe the bug

The Docker image does not contain the app version data, causing some links to break.

The root issue is that the Makefile depends on a git command to get the rev, and the Docker builder ignores the .git folder. You can see the error here: https://github.com/writefreely/writefreely/actions/runs/7758253252/job/21159589862#step:5:365

I recommend adding a fallback to an git rev environment variable that is set in the Docker build action.

1. Run in the official Docker image
1. Go to https://example.com/admin
1. Observe the footer links to <https://writefreely.org/guide/v> and says "writefreely v"

### Expected behavior
It should say the version number in the footer

### Application configuration
 - **Single mode** 
 - **Database?** sqlite
 - **Open registration?** no
 - **Federation enabled?** yes

**Version or last commit**: ghcr.io/writefreely/writefreely:v0.15.0

claabs:
- Include .git context in Docker build for UI version number
  - Fixes #873
  - Potential caching/performance impact, but it's a really simple solution that doesn't require any special scripting that would break local Docker Compose dev usage and whatnot.

---

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,2 +1,1 @@
 Dockerfile
-.git


thebaer:
Thanks for fixing this, and bringing up the healthcheck issue!

Just as an FYI, we prefer each PR to address only one issue at a time. Like in this case, it seems the healthcheck issue was also addressed in a different way in #875. I'm fine with either solution, but for this PR, could you remove the changes in https://github.com/writefreely/writefreely/pull/874/commits/9e14b5ffd1684587e8dddf374151b3671df5eff1 and then either open an individual PR for it, or let us know if you think #875 is a fine solution for the problem. With that sorted out, we can get this merged!

thebaer (approved):

PR #875, opened 2024-02-07
Add healthcheck dependency

kotnik:
Health check from `Dockerfile` uses `curl` but it is not installed by default. This causes container to be forever unhealthy and thus ignored by other apps, for example Traefik.

---


kotnik:
Possible dupe of #874, but chooses to keep `curl` instead of switching to `wget`.

thebaer:
Appreciate you taking a look at this! I don't have a huge opinion on which route is better, but I think we'll go with #894 just because it doesn't add another dependency.

If you want to help with other [Docker-related issues](https://github.com/writefreely/writefreely/labels/docker), you're more than welcome to!

PR #883, opened 2024-02-15
add f.created to join, add Created to Scan

elkcityhazard:
resubmitting fix for "subscribed since" null time bug.  

fixed issue with go versioning and formatting.

obtain f.created from join statement and pass into the value pointed at by RemoteUser argument.  

---

### DIFF
--- a/database.go
+++ b/database.go
@@ -14,14 +14,15 @@ import (
 	"context"
 	"database/sql"
 	"fmt"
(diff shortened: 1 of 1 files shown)


snullp (approved):

thebaer (approved):
Looks good! Thanks, merging now.

PR #894, opened 2024-02-20
Fix Docker healthcheck with wget

claabs:
- Change Docker healthcheck to `wget`
  - No issue open for this one, but the healthcheck fails because `curl` is not available on the image. Switched to `wget` and the best practice args for a healthcheck
  - Alternative to #875 that doesn't install any new packages on the image

---


thebaer:
I think we'll go with this one, since it doesn't add another dependency. Thanks again for contributing!

PR #948, opened 2024-03-09
add support work over proxy HTTP/HTTPS/SOCKS

iwojim0:
https://github.com/writefreely/writefreely/issues/947
 - added possibility use http_proxy environment variable
 - for overlay networks also need  increase timeout for http requests


thebaer:
Thanks for submitting this! My main concern here is that each outgoing request to an ActivityPub server happens in sequence, so any slow server will hold up other requests as they're going out to the fediverse. With this change, it could be even slower than before.

I think to resolve this, we'd need to either change how requests go out (i.e. send them in parallel), or make this configurable somehow.

PR #1038, opened 2024-09-02
Docker for Production

leo9800:
This patch contains necessary changes to enable users to run writefreely in docker for production purposes.

Besides, a sample `docker-compose.yml` file, named `docker-compose.prod.yml`, is provided with MariaDB as database.

### Full sample procedure for setup

1. `cd` to directory of writefreely repository, check out `docker-prod` branch
2. build docker image by `docker build -t writefreely .`
3. create a directory, e.g. `/srv/docker/writefreely`
4. copy `docker-compose.prod.yml` to `/srv/docker/writefreely`
5. `cd /srv/docker/writefreely`
6. `docker compose run -it --rm app writefreely config start` and run normal setup procedure, **choose "Production, behind reverse proxy" mode**, input database name, username and password specified in `docker-compose.yml`, set database host to `db`.
7. `docker compose run -it --rm app writefreely keys generate` to generate keys
8. `docker compose up -d`
9. checkout http://localhost:8080 with web browser

Besides, a sample reverse proxy configuration with Apache HTTP Server is available as below:

```
<VirtualHost *:443>
	ServerName domain.tld
	SSLEngine On
	SSLCertificateFile /path/to/ssl.crt
	SSLCertificateKeyFile /path/to/ssl.key

	ProxyPass / "http://127.0.0.1:8080/" nocanon
	ProxyAddHeaders Off
	ProxyPreserveHost On
	ProxyErrorOverride Off

	RequestHeader set "X-Real-IP" expr=%{REMOTE_ADDR}
	RequestHeader set "X-Forwarded-For" expr=%{REMOTE_ADDR}
	RequestHeader set "X-Forwarded-Port" expr=%{SERVER_PORT}
	RequestHeader set "X-Forwarded-Proto" expr=%{REQUEST_SCHEME}
	RequestHeader set "X-Forwarded-Ssl" expr=%{HTTPS}

	Header unset server
</VirtualHost>
```

I am running this setup in my production server, thus, I would like to share it with the community. :-D

---

### DIFF
--- a/config/setup.go
+++ b/config/setup.go
@@ -12,12 +12,14 @@ package config
 
 import (
 	"fmt"
+	"os"
+	"strconv"
+	"strings"
+
 	"github.com/fatih/color"
 	"github.com/manifoldco/promptui"
 	"github.com/mitchellh/go-wordwrap"
 	"github.com/writeas/web-core/auth"
-	"strconv"
-	"strings"
 )
 
 type SetupData struct {
@@ -79,6 +81,8 @@ func Configure(fname string, configSections string) (*SetupData, error) {
 		}
 		isDevEnv := envType == "Development"
 		isStandalone := envType == "Production, standalone"
+
+		_, isDocker := os.LookupEnv("WRITEFREELY_DOCKER")
 
 		data.Config.Server.Dev = isDevEnv
 
@@ -148,6 +152,16 @@ func Configure(fname string, configSections string) (*SetupData, error) {
 		} else {
 			data.Config.Server.TLSCertPath = ""
 			data.Config.Server.TLSKeyPath = ""
+		}
+
+		// If running in docker:
+		// 1. always bind to 0.0.0.0 instead of localhost
+		// 2. set paths of static files in UNIX manners
+		if !isDevEnv && isDocker {
+			data.Config.Server.TemplatesParentDir = "/usr/share/writefreely"
+			data.Config.Server.StaticParentDir = "/usr/share/writefreely"
+			data.Config.Server.PagesParentDir = "/usr/share/writefreely"
+			data.Config.Server.Bind = "0.0.0.0"
 		}
 
 		fmt.Println()
--- a/Dockerfile.prod
+++ b/Dockerfile.prod
@@ -1,0 +1,34 @@
+FROM golang:alpine AS build
+
+LABEL org.opencontainers.image.source="https://github.com/writefreely/writefreely"
+LABEL org.opencontainers.image.description="WriteFreely is a clean, minimalist publishing platform made for writers. Start a blog, share knowledge within your organization, or build a community around the shared act of writing."
+
+RUN apk update --no-cache && \
+	apk upgrade --no-cache && \
+	apk add --no-cache nodejs npm make g++ git sqlite-dev patch && \
+	npm install -g less less-plugin-clean-css && \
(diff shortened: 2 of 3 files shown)


config/setup.go
```diff
@@ -150,6 +154,16 @@ func Configure(fname string, configSections string) (*SetupData, error) {
...
+		// If running in docker:
+		// 1. always bind to 0.0.0.0 instead of localhost
+		// 2. set paths of static files in UNIX manners
+		if !isDevEnv && isDocker {
+			data.Config.Server.TemplatesParentDir = "/usr/share/writefreely"
+			data.Config.Server.StaticParentDir = "/usr/share/writefreely"
+			data.Config.Server.PagesParentDir = "/usr/share/writefreely"
+			data.Config.Server.Bind = "0.0.0.0"
+		}
+
```
thebaer, line 166: Could we use a Docker-specific configuration file instead, or do you think this is the most straightforward way to way to handle this?

thebaer (commented):
Thank you so much for submitting this! Just one question on this, to be sure we're not adding too much custom code when it can potentially be handled in the config.

config/setup.go
leo9800: I actually tried to do so by creating a template `config.ini` in `/data` directory inside the container with following content:

```
[server]
bind                 = 0.0.0.0
templates_parent_dir = /usr/share/writefreely
static_parent_dir    = /usr/share/writefreely
pages_parent_dir     = /usr/share/writefreely
```

which is equivalent to the lines added to `config/setup.go`.

But when running `writefreely config start`, writefreely seems read the (incomplete) `config.ini` file, and results in all the default values (e.g. DB filename `writefreely.db` if the user choice sqlite) omitted. **This would not provide an uniform user experience compared with non-docker setup**.

Thus, in my opinion, adding these lines may be the solution with the least modification to source code I could come up with.

Besides, these 4 lines are only effective when envvar `WRITEFREELY_DOCKER` is set, so probably incurs no impact on existing non-docker installs?

thebaer: Yep, that makes sense. Thanks for explaining!

thebaer (approved):
Looks good to me! I'll merge this now.

We'll also need to update our [documentation](https://github.com/writefreely/documentation) with this change. If you're able to contribute, specifically to [this file](https://github.com/writefreely/documentation/blob/main/admin/docker.md), would really appreciate that!

carbontwelve:
@leo9800 where does `config.ini` and the `keys` folder end up getting created?

leo9800:
@carbontwelve They should be stored within `/data` inside the docker container,

Given that container path `/data` should be mapped to host (as a docker volume), the path of `config.ini` and `keys` depends on your `docker run` command or `docker-compose.yml` stack configuration.

e.g. if you are working with a `docker-compose.yml` looks like:
```
services:
  app:
    image: writefreely
    container_name: writefreely
    volumes:
      - /path/to/writefreely:/data
    # omitted
```
then try finding them under directory `/path/to/writefreely` on your host.

carbontwelve:
You're right @leo9800. 

My confusion was that the container I used is the one available here  https://github.com/writefreely/writefreely/pkgs/container/writefreely which is built from the original `Dockerfile` and not the `Dockerfile.prod`.

leo9800:
@carbontwelve
Unfortunately, according to https://writefreely.org/docs/main/admin/docker the docker image you used is for developing purposes, not for production deployment…

PR #1047, opened 2024-09-27
Switch to gosimple/slug from writeas/slug

thebaer:
The upstream library now has the changes we'd merged in this fork, so no need to use our (very outdated) fork anymore.

---


thebaer:
Merging now.

PR #1103, opened 2024-10-17
Fix broken links in drafts loaded with "load more" #1054

### LINKED ISSUE #1054: drafts loaded with "load more" have broken links
1. create ~13 drafts
2. go to /me/posts/
3. click on "Load More.."

the loaded post's "edit" button will link to /pad/POSTID instead of /POSTID/edit, and the link may say "edited" instead of "edit".

### Application configuration
 - **Single mode or Multi-user mode?**: multi-user
 - **Database?**: unknown
 - **Open registration?** yes
 - **Federation enabled?** yes

**Version or last commit**: [writefreely v0.13.2](https://writefreely.org/)

vtyeh:
### Description
This PR fixes broken edit links for drafts when loaded via 'load more' by dynamically generating the edit links based on the `single_user` config value.

Previously, drafts loaded with 'load more' had broken links because the link generation didn't account for the single-user mode and the path /pad/ seems to be deprecated.

Original issue description: [Issue #1054](https://github.com/writefreely/writefreely/issues/1054).

### Changes Made
- In `posts.js`, I introduced a new argument of `singleUser` to the `createPostEl` method
- In `articles.tmpl`, I pass down the single_user boolean to the `createPostEl` method

### How to Test
To verify the edit links work correctly in both single-user and multi-user modes:

1. Ensure you have more than 10 drafts
2. Set `single_user` = `true` in config.ini
3. Restart the app
4. Go to Drafts and click "Load more..."
5. Click "Edit" on one of the loaded drafts:
    - In single-user mode, the link should look like `/d/POSTID/edit`
    - In multi-user mode, it should look like `/POSTID/edit`
6. Set `single_user` = `false` in config.ini
7. Restart the app and repeat steps 4-5

### Related Ticket
https://github.com/writefreely/writefreely/issues/1054

### Additional Notes
I tried to see if introducing the new argument would cause any issues when the method is called elsewhere. I found that the other usages are in `posts.js` under `delPost` method. I am not sure how to test the functionality under the comment "Fill in full page of posts" to see if it is adversely affected by the introduction of the new method argument.

I found another minor bug while testing this and will open a separate issue for it. A quick summary: When deleting drafts, it doesn't refill the 10 latest drafts, so you can delete all 10 until all that is left is the Load more... link. Once you click it, the remaining drafts populate.

---

### DIFF
--- a/static/js/posts.js
+++ b/static/js/posts.js
@@ -193,16 +193,17 @@ var createPostEl = function(post, owned, singleUser) {
 	let p = H.createPost(post.id, "", post.body)
 	var title = (post.title || p.title || post.id);
 	title = title.replace(/</g, "&lt;");
+	var postLink = (singleUser ? '/d/' : '/') + post.id + '/edit'
 	$post.id = 'post-' + post.id;
 	$post.className = 'post';
-	$post.innerHTML = '<h3><a href="/' + post.id + '">' + title + '</a></h3>';
+	$post.innerHTML = '<h3><a href="' + postLink + '">' + title + '</a></h3>';
 
 	var posted = "";
 	if (post.created) {
 		posted = getFormattedDate(new Date(post.created))
 	}
 	var hasDraft = H.exists('draft' + post.id);
-	$post.innerHTML += '<h4><date>' + posted + '</date> <a class="action" href="' + (singleUser ? '/d/' : '/') + post.id + '/edit">edit' + (hasDraft ? 'ed' : '') + '</a> <a class="delete action" href="/' + post.id + '" onclick="delPost(event, \'' + post.id + '\'' + (owned === true ? ', true' : '') + ')">delete</a> '+movePostHTML(post.id)+'</h4>';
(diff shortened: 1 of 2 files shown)


thebaer:
Thanks for fixing this! Everything looks good at first glance, but I'll test shortly.

Please feel free to open a new issue for filling in more posts as you delete posts. I believe I have a fix ready to go for that already.

vtyeh:
Thank you!

I opened the [new issue](https://github.com/writefreely/writefreely/issues/1123).

thebaer (changes requested):
This works perfectly!

Just one additional bug I noticed that I think we should fix here: loaded post titles (L198 in posts.js) should also have the condition applied. Single-user instances should link to `/d/{id}` and multi-user instances should still link to `/{id}`.

vtyeh:
Nice catch! I pulled out the link into its own variable since we use it multiple times. Just pushed up the new change.

thebaer:
Ah! That title link should actually go directly to the post page (without the `/edit` suffix), while the edit link should still go to `/edit`. Sorry for the confusion!

vtyeh:
Ah, no worries! That makes a lot of sense. I've pushed up the new change.

thebaer (approved):
Perfect! This looks great. Thanks again for contributing!

PR #1116, opened 2024-10-21
Fix glibc issue

### LINKED ISSUE #1064: WriteFreely won't start on older OSes due to glibc issue
### Describe the bug

When attempting to run the WriteFreely v0.15.1 release build on anything before Ubuntu 20.04, it fails to start with this error message:

```
./writefreely: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.28' not found (required by ./writefreely)
```

This didn't happen with Go 1.19, but with the update to Go 1.21 (#1025), we now see this on systems that don't have this version of `glibc` (see golang/go#58550). We can't use the `CGO_ENABLED=0` fix, because we need CGO for sqlite.

Instead, we should statically compile WriteFreely for release -- see https://github.com/golang/go/issues/26492#issuecomment-435462350 for a potential starting point.

### Workaround

Admins can compile WriteFreely from scratch on their target server by cloning the repository and running:

```
make build
```

### Application configuration
 - **Single mode or Multi-user mode?** n/a
 - **Database?** n/a
 - **Open registration?** n/a
 - **Federation enabled?** n/a

**Version or last commit**: v0.15.1 (release build)

sahilmulla:
Fix for #1064

---

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 GITREV=`git describe | cut -c 2-`
-LDFLAGS=-ldflags="-s -w -X 'github.com/writefreely/writefreely.softwareVer=$(GITREV)'"
+LDFLAGS=-ldflags="-s -w -X 'github.com/writefreely/writefreely.softwareVer=$(GITREV)' -extldflags '-static'"
 
 GOCMD=go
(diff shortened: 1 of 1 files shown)


thebaer:
Thanks for fixing! This looks great -- merging now.

PR #1127, opened 2024-10-22
Fix two error messages: user create and user create --admin

DonPiotr:
I'm not GO dev, but at least I can fix some messages.

---


thebaer (approved):
Thanks for contributing this! Looks great, will merge now.

PR #1133, opened 2024-10-28
support go task

sahilmulla:
Adding support for go task

---


thebaer:
Thanks for submitting this. Can you share a little more about how this should be used / how it integrates with our current workflows?

PR #1136, opened 2024-10-28
Docker setup improvement: Run WriteFreely with Docker in production

jpavonabian:
This PR was opened following the discussion in the forum post [Run WriteFreely with Docker.](https://discuss.write.as/t/run-writefreely-with-docker/14071/4)
Updated the build workflow to create separate Docker images for production and development  environments:
- The production image is built and tagged when a new version tag is pushed (e.g., v1.0.0), ensuring that stable releases are clearly defined.
- Development images are created when changes are pushed to the main or develop branches, allowing for ongoing development without affecting the production environment.

Created a New Script (docker_setup_prod.sh):
This script automates the setup of WriteFreely in a production environment. It downloads the necessary docker-compose file and guides the user through the configuration process.

---


MGessinger:
Surely I must be missing something. Following the instructions in #1038 , I was able to get up and running right away, but using the script from this branch, the configuration step fails because it tries to read config.ini on startup (which of course doesn't exist yet).
Am I doing something wrong or is the script doing something wrong? Is this perhaps due to the fact that on this branch I am pulling from the repo, wheras the instructions in #1038 build locally?

MGessinger:
After some more investigation, I am certain that it must have been my mistake, since the script doesn't do much other than fetching the docker compose and running the same config commands as the linked PR does.

However, in my debugging, I came across some questions that relate to this PR specifically, so if I may, I will collect them below.

- Part of the script is to fetch a docker compose file. Ideally, after this PR is merged, that won't be necessary and the docker-compose file should be part of the repository, right? In fact, it may lead to compatibility issues to always fetch the latest docker-compose from the remote repo, since the latest version may contain breaking changes when compared to the version that is currently checked out (and from which the docker image was built).
- In the same vein as the previous comment, it might make sense to include the tagged version number in the docker-compose file, so that when I check out a particular commit, the script will always pull and run the correct version of the docker image, rather than latest.

jpavonabian:
> After some more investigation, I am certain that it must have been my mistake, since the script doesn't do much other than fetching the docker compose and running the same config commands as the linked PR does.
> 
> However, in my debugging, I came across some questions that relate to this PR specifically, so if I may, I will collect them below.
> 
>     * Part of the script is to fetch a docker compose file. Ideally, after this PR is merged, that won't be necessary and the docker-compose file should be part of the repository, right? In fact, it may lead to compatibility issues to always fetch the latest docker-compose from the remote repo, since the latest version may contain breaking changes when compared to the version that is currently checked out (and from which the docker image was built).
> 
>     * In the same vein as the previous comment, it might make sense to include the tagged version number in the docker-compose file, so that when I check out a particular commit, the script will always pull and run the correct version of the docker image, rather than latest.
Hi,
I am not clear if my answers were sent, as I responded by e-mail and I do not see them. Anyway, I will reply again just in case they were missed.
I tried to get the Workflow (./github/workflows/docker-publish-pro.yml) to update the docker-compose, but I wasn't able to. You can always before creating a tag update the file by hand. It would certainly be ideal to quickly return to a specific version.
You could certainly clone the repository and not use the script. The script is intended to help people deploy easily, but you have no need to use it at all.
You create a directory, get the docker-compose via the script and follow the instructions. Something like this: curl -o docker_setup_prod.sh https://raw.githubusercontent.com/writefreely/writefreely/refs/heads/develop/docker_setup_prod.sh && bash docker_setup_prod.sh
You can forgo the script if you wish. Simply use the docker-compose.prod.yml file that you will find in the repository to launch the instance. It is recommended to place the file in a specific directory for the WriteFreely instance and rename it to docker-compose.yml for easier management.

Once that is done, you just need to run the commands
```
docker compose run -it --rm app writefreely config start
 docker compose run -it --rm app writefreely keys generate
 ```
 before bringing up the containers with
 ```
 docker-compose up -d
```

MGessinger:
Just to avoid any misunderstandings, I was not trying to imply that this script is not useful. I think it is a valuable addition to the repository to have a quick and easy way to get a production image up and running with just one command. The only reason it didn't work for me is because I used it wrong.

For anyone else reading this thread, the mistake I made was to use the updated docker-compose file which is part of this PR. This updated file refers to the docker image in the github registry. But at the time of writing this, the github registry does not yet contain the prod build of the docker image. Once this PR is merged and the pipelines have run, that won't be an issue anymore and the script should work fine.

Coming back to the conversation at hand, it was this incompatibility that prompted me to post the above comment. The setup script assumes that the docker image behaves in a particular way, but it can't know. It also doesn't refetch the docker-compose file if it already exists. Both of these will make it very hard to debug docker issues down the line. Ideally, the setup script and the docker-compose file should be shipped out as a bundle, so that compatibility can be ensured.
For the same reason, I think the version numbers should be pinned to exact values. I understand that it comes with additional manual work to adjust version numbers everywhere, but it comes with the added benefit of documenting your breaking changes into the repository.
The tagged commit would then be the one which bumps version numbers everywhere.

jpavonabian:
> Just to avoid any misunderstandings, I was not trying to imply that this script is not useful. I think it is a valuable addition to the repository to have a quick and eady way to get a production image up and running with just one command. The only reason it didn't work for me is because I used it wrong.
> 
> For anyone else reading this thread, the mistake I made was to use the updated docker-compose file which is part of this PR. This updated file refers to the docker image in the github registry. But at the time of writing this, the github registry does not yet contain the prod build of the docker image. Once this PR is merged and the pipelines have run, that won't be an ossue anymore and the script should work fine.
> 
> Coming back to the conversation at hand, it was this incompatibility that prompted me to post the above comment. The setup script assumes that the docker image behaves in a particular way, but it can't know. It also doesn't refetch the docker-compose file if it already exists. Bith of these will make it very hard to debug docker issues down the line. Ideally, the setup script and the docker-compose file should be shipped out as a bundle, so that compatibility can be ensured. For the same reason, I think the version numbers should be pinned to exact values. I understand that it xomes with additional manual work to adjust version numbers everywhere, but it comes with the added benefit of documenting your breaking changes into the repository. The tagged commit would then be the one which bumps version numbers everywhere.

I understood that there was no bad intention on your part, no problem.

PR #1137, opened 2024-10-29
Fix date format

sahilmulla:
Fix for #866

---


thebaer:
Thanks for taking a look at this! However I believe, just from looking at the code, that this will break how dates get updated from the "Update post" page on the web. I think we'll need to fix that on the frontend before we can fix this in the API.

I'm going to tackle the frontend side, including upgrading the field to a `datetime-local` field, and then we should be able to use this change.

thebaer:
I just finished this work up in #1492, so I'll go ahead and close this out. Would appreciate any feedback you might have there, if you're interested! Thanks again.

PR #1141, opened 2024-11-08
Added mastodon author attribution support

JonathanS223:
Adding functionality to template so that posts are able to display the Mastodon Author Attribution for Mastodon 4.3

---

- [X] I have signed the [CLA](https://todo.musing.studio/L1)

### DIFF
--- a/templates/chorus-collection.tmpl
+++ b/templates/chorus-collection.tmpl
@@ -42,7 +42,7 @@
     }
 
     // URL to be converted
-    const url = "https://social.vivaldi.net/@ItWasntMe223"; 
+    const url = "{{.Verification}}"; 
 
     // Convert the URL
     const convertedUrl = convertUrl(url);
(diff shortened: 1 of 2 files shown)


thebaer:
Thanks for contributing this! Just to be sure, have you confirmed that this works with the current system we have? That is, just using the fediverse profile _URL_ as the `content` value, instead of the profile _handle_? Last I saw, it looked like it needed a handle there.

JonathanS223:
@thebaer,

Gah! You are absolutely right. It's only using the handle. I need to see if we have the handle. Back to the drawing board.

thebaer:
No worries! I have to verify, but I believe we do store the handle in the `remoteusers` table. Maybe we could use that? Also, note that any client-side / JS solution probably won't work, as Mastodon only checks generated HTML when verifying.

JonathanS223:
Ahh. Gotcha. I was going through trying to find where we stored it cause I didn't want to force us to have to create a new entry in the db. I was thinking of taking the handle and twisting it around with JS to make it match what is should. 

I'll dig there where you recommend. Thanks!

thebaer:
Hey, again appreciate you taking a look at this! But I've just opened #1253 to solve this -- it also pulls the handle from the local database to make sure it works with all fediverse platforms. Please feel free to review and give input, or look at any other issues! For now, I'll close this.

JonathanS223:
Not a problem @thebaer . I got taken down by Covid so I never got to touch this again. Glad to see it was added!

PR #1142, opened 2024-11-11
display stat numbers as monospace

lolbinarycat:
this allows you to reliably compare the magnitude of numbers just by looking at their width.

before:
image

after:
image

---


thebaer:
Sorry for the slow response, and thanks for fixing this! Looks good, merging now.

PR #1145, opened 2024-11-21
Increased `posts.content` size for MySQL.

### LINKED ISSUE #987: Limited post size on MySQL backend
### Describe the bug

`posts.content` is modelled as a `text` column [on `schema.sql`](https://github.com/writefreely/writefreely/blob/3a6118c2070344fdc74a5918aeba1dc55913e965/schema.sql#L139).

The `TEXT` data type [has a limit of 2^16=64KB on MySQL](https://dev.mysql.com/doc/refman/8.0/en/storage-requirements.html).

This means that longer posts (longer than ~8000/9000 words in a language like English) can't be stored without having the database to throw an error.

(Note that you don't have this limitation on SQLite, where the maximum size of a data type isn't enforced and it only obeys the laws of `SQLITE_MAX_LENGTH`, which is usually ~1GB).

1. Create a post with more than 64K characters on a MySQL backend
2. Submit
3. Submission will fail - a database error will be logged

### Expected behavior
Posts up to a reasonable length for a blog article should be allowed on all the supported databases.

### Application configuration
 - **Single mode or Multi-user mode?** 
 - **Database?** mysql
 - **Open registration?** N/A
 - **Federation enabled?** N/A

**Version or last commit**:
N/A

**Workarounds**

1. If you are planning to initialize your db through : change the data type of [the `content` column on the `posts` table](https://github.com/writefreely/writefreely/blob/3a6118c2070344fdc74a5918aeba1dc55913e965/schema.sql#L139) - from `text` to `mediumtext`.
2. If you have an existing database, change the data type of `posts.content`:

```sql
ALTER TABLE posts MODIFY COLUMN content MEDIUMTEXT CHARACTER SET utf8mb4 COLLATE utf8mb4_bin NOT NULL;
```

**Proposal**

Change `posts.content` data type from `TEXT` (size: 2^16 on MySQL) to `MEDIUMTEXT` (size: 2^24 on MySQL).

It's quite straightforward how to do it from `schema.sql` or via `alter` commands, but I'm not familiar enough with the [SQL generation logic in `db/create.go`](https://github.com/writefreely/writefreely/blob/develop/db/create.go) (I could take a look into it if there's interest in a PR though). At a first glance it seems that [a new `ColumnTypeMediumText` needs to be added](https://github.com/writefreely/writefreely/blob/78e59b749b003f39053d82d001a2b91bebb706ee/db/create.go#L59) and then [handled](https://github.com/writefreely/writefreely/blob/78e59b749b003f39053d82d001a2b91bebb706ee/db/create.go#L59) (as `MEDIUMTEXT` for MySQL, and simple `TEXT` for SQLite), but I'm not sure if this will cover everything (plus I guess that there would also be a migration logic to add).

blacklight:
`TEXT` -> 64kB
`MEDIUMTEXT` -> 16MB

Closes: https://github.com/writefreely/writefreely/issues/987

---

### DIFF
--- a/schema.sql
+++ b/schema.sql
@@ -136,7 +136,7 @@ CREATE TABLE IF NOT EXISTS `posts` (
   `updated` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
   `view_count` int(6) NOT NULL,
   `title` varchar(160) CHARACTER SET utf8mb4 COLLATE utf8mb4_bin NOT NULL,
-  `content` text CHARACTER SET utf8mb4 COLLATE utf8mb4_bin NOT NULL,
+  `content` mediumtext CHARACTER SET utf8mb4 COLLATE utf8mb4_bin NOT NULL,
   PRIMARY KEY (`id`),
   UNIQUE KEY `id_slug` (`collection_id`,`slug`),
   UNIQUE KEY `owner_id` (`owner_id`,`id`),
--- a/schema.sql
+++ b/schema.sql
@@ -136,7 +136,7 @@ CREATE TABLE IF NOT EXISTS `posts` (
   `updated` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
   `view_count` int(6) NOT NULL,
   `title` varchar(160) CHARACTER SET utf8mb4 COLLATE utf8mb4_bin NOT NULL,
-  `content` mediumtext CHARACTER SET utf8mb4 COLLATE utf8mb4_bin NOT NULL,
+  `content` text CHARACTER SET utf8mb4 COLLATE utf8mb4_bin NOT NULL,
   PRIMARY KEY (`id`),
   UNIQUE KEY `id_slug` (`collection_id`,`slug`),
   UNIQUE KEY `owner_id` (`owner_id`,`id`),
--- a/db/create.go
+++ b/db/create.go
@@ -57,6 +57,7 @@ const (
 	ColumnTypeChar     ColumnType = iota
 	ColumnTypeVarChar  ColumnType = iota
 	ColumnTypeText     ColumnType = iota
+	ColumnTypeLongText ColumnType = iota
 	ColumnTypeDateTime ColumnType = iota
 )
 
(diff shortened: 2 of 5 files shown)


blacklight:
Any plans to review/merge this? There's now a conflict because the v16 migration file has already been taken. I can quickly fix it but only if there's a plan to actually merge it (otherwise it's going to be a catch-up-with-db-migrations game).

thebaer:
Yep, happy to resolve this -- and thanks for contributing a fix! I'm a little concerned this might allow people to stuff a bunch of data into posts and fill up the database quickly. But that's probably not a major issue we need to address right now. Please feel free to fix this, and we'll get this merged. I'll also review the code shortly.

schema.sql
```diff
@@ -136,7 +136,7 @@ CREATE TABLE IF NOT EXISTS `posts` (
   `updated` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
   `view_count` int(6) NOT NULL,
   `title` varchar(160) CHARACTER SET utf8mb4 COLLATE utf8mb4_bin NOT NULL,
-  `content` text CHARACTER SET utf8mb4 COLLATE utf8mb4_bin NOT NULL,
+  `content` mediumtext CHARACTER SET utf8mb4 COLLATE utf8mb4_bin NOT NULL,
```
thebaer: This change isn't needed -- `schema.sql` only represents the initial state of the database, and then all migrations are applied over it.

thebaer (changes requested):
Thanks again for addressing this! Besides fixing the merge conflict, just one note here, and then you'll also want to add the migration function call to `migrations/migrations.go` in order for this to run.

schema.sql
blacklight: @thebaer sorry for the delayed response. I have addressed this and updated the PR accordingly.

thebaer (changes requested):
Thanks for updating this -- still one unaddressed issue:

> you'll also want to add the migration function call to `migrations/migrations.go` in order for this to run.

Right now, this migration never actually runs.

blacklight:
Any plans to merge this at some point?

blacklight:
This has been waiting to merge for too long. I'll close it because I'm not keen to keep it up-to-date with new db schema changes being merged before it. Feel free to reopen it and merge it if interested.

PR #1153, opened 2024-12-04
Fixes for smtp branch

jbgi:
I've tested this on my WF server. Emails now working with smtp.
~~**Except Unsubscribe link** in email which is still raw template. Would need some refactoring to fix this.~~ (fixed in last commit)

---

### DIFF
--- a/app.go
+++ b/app.go
@@ -428,15 +428,11 @@ func Initialize(apper Apper, debug bool) (*App, error) {
 
 	initActivityPub(apper.App())
 
-	if apper.App().cfg.Email.Domain != "" || apper.App().cfg.Email.MailgunPrivate != "" {
-		if apper.App().cfg.Email.Domain == "" {
-			log.Error("[FAILED] Starting publish jobs queue: no [letters]domain config value set.")
-		} else if apper.App().cfg.Email.MailgunPrivate == "" {
-			log.Error("[FAILED] Starting publish jobs queue: no [letters]mailgun_private config value set.")
-		} else {
-			log.Info("Starting publish jobs queue...")
-			go startPublishJobsQueue(apper.App())
-		}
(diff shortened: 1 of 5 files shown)


app.go
```diff
@@ -428,15 +428,11 @@ func Initialize(apper Apper, debug bool) (*App, error) {
...
-	if apper.App().cfg.Email.Domain != "" || apper.App().cfg.Email.MailgunPrivate != "" {
-		if apper.App().cfg.Email.Domain == "" {
-			log.Error("[FAILED] Starting publish jobs queue: no [letters]domain config value set.")
-		} else if apper.App().cfg.Email.MailgunPrivate == "" {
-			log.Error("[FAILED] Starting publish jobs queue: no [letters]mailgun_private config value set.")
-		} else {
-			log.Info("Starting publish jobs queue...")
-			go startPublishJobsQueue(apper.App())
-		}
+	if apper.App().cfg.Email.Enabled() {
```
thebaer, line 431: We may revisit this to add more helpful config error messages

mailer/mailer.go
```diff
@@ -58,14 +76,16 @@ func (m *Mailer) NewMessage(from, subject, text string, to ...string) (*Message,
...
-
-		if msg.smtpMsg.Error != nil {
-			return nil, msg.smtpMsg.Error
+		msg.smtpMsg = &SmtpMessage {
+			from,
+			"",
+			subject,
+			make([]Recipient, len(to)),
+			"",
+			text,
```
thebaer: We should add in field names here, and run `go fmt`

thebaer (commented):
This looks great! Thanks so much for helping out.

I left a few comments we might address in the future, but no need to make those changes now if you don't want. I'll get this merged shortly.

thebaer:
Thanks again -- merging now

PR #1253, opened 2025-02-15
Display fediverse:creator header tag on posts

thebaer:
This uses the fediverse actor a blog author has added in the Verification field on their blog, looking up the handle locally (required by this tag, instead of the URL required for the `rel=me` tag) to display it in the `fediverse:creator` tag. If it doesn't exist locally, the tag is simply left out.

**Work in progress**, because ideally we'd always show the `fediverse:creator` tag. Right now, this will work if someone inputs their handle (e.g. `@matt@writing.exchange`) in the Verification field, because when saved, we fetch the remote user. We could solve this by _always_ fetching the remote user when updating the Verification value.

**Update**: see this comment: https://github.com/writefreely/writefreely/pull/1253#issuecomment-3221587222 and [T932](https://todo.musing.studio/T932) for remaining work.

---

### DIFF
--- a/activitypub.go
+++ b/activitypub.go
@@ -971,6 +971,24 @@ func getRemoteUserFromHandle(app *App, handle string) (*RemoteUser, error) {
 	return &u, nil
 }
 
+// getRemoteUserFromURL retrieves the profile page of a remote user
+// from the @user@server.tld handle
+func getRemoteUserFromURL(app *App, urlStr string) (*RemoteUser, error) {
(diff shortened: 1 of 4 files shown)


thebaer:
Going to include this in v0.16 as "good enough," with the caveat that users should input their @-handle in the Verification field for now to ensure this works. Otherwise, the remaining bit of work will be fixed with [T932](https://todo.musing.studio/T932).

PR #1254, opened 2025-02-17
Changed the permission of config file to allow access to only owner. …

AryanDame:
Changed the permission of config file to allow access to only owner. This fixes CVE-2025-24337

Changed the permission of config file to allow access to only owner. This fixes CVE-2025-24337
---

### DIFF
--- a/app.go
+++ b/app.go
@@ -554,7 +554,7 @@ requests. We recommend supplying a valid host name.`)
 		}
 
 		if network == "unix" {
-			err = os.Chmod(bindAddress, 0o666)
+			err = os.Chmod(bindAddress, 0o600)
 			if err != nil {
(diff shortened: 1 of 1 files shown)


thebaer:
Thanks for taking a look at this, but it seems this change sets the permissions of the _socket_ file (if WF is configured to use that), instead of permissions on the _config_ file.

thebaer (changes requested):
Thanks for taking a look at this, but it seems this change sets the permissions of the _socket_ file (if WF is configured to use that), instead of permissions on the _config_ file.

thebaer:
Closing this since there hasn't been any movement on it, and the issue should be fixed correctly now through separate changes.

PR #1366, opened 2025-05-03
Fix like counts on stats page

ltning:
Fix the status page SQL to include like counts for posts.

---

### DIFF
--- a/database.go
+++ b/database.go
@@ -2044,7 +2044,7 @@ func (db *datastore) GetTopPosts(u *User, alias string, hostName string) (*[]Pub
 		where = " AND alias = ?"
 		params = append(params, alias)
 	}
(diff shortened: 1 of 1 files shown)


thebaer:
Hey, thanks for bringing up this bug and fixing! It looks like we were just missing a few lines that should've been in there from the start, which I've included in #1459. We'll go with the additional query there, instead of using another `LEFT JOIN`, as I believe that will be a little better performance-wise. But thanks again for contributing!

ltning:
> Hey, thanks for bringing up this bug and fixing! It looks like we were just missing a few lines that should've been in there from the start, which I've included in #1459. We'll go with the additional query there, instead of using another `LEFT JOIN`, as I believe that will be a little better performance-wise. But thanks again for contributing!

Hola, cool. I don't quite see why you'd want another query though - a join like this is very cheap and the query plan looked very sane when I checked. I tried to simulate with a lot more posts++ than I have in my instance, and the filtering remains efficient.

My experience is that if a join is clean it's always the fastest approach. The overhead of submitting, analyzing, planning, executing and returning data for a second query will pretty much always be much much worse.

PR #1372, opened 2025-05-07
Fix security context getting removed on AP collection endpoint

thebaer:
Previously, we were overwriting the `context` property when fetching a collection as an ActivityPub object. This fixes that.

[Originally reported by @js@podcastindex.social](https://podcastindex.social/@js/113041934101924940)

---


thebaer:
Merging now!

PR #1373, opened 2025-05-08
Fix ActivityPub following from Ghost

thebaer:
This makes Follow request parsing more robust. Previously, this only worked if the `object` was a URI, which is what many platforms send. Now, we can also handle objects here.

---


thebaer:
Working well in production, merging now.

PR #1374, opened 2025-05-08
Add `preview` property to Articles

thebaer:
This adds a `preview` property (a fallback `Note`) on each ActivityPub `Article` we send out. Other fediverse platforms that support it can use this object, instead. It aims to adhere to [FEP-b2b8: Long-form Text](https://codeberg.org/fediverse/fep/src/branch/main/fep/b2b8/fep-b2b8.md).

This implementation aims to be as automatic and non-intrusive to the user as possible. It first searches for a user-defined excerpt, marked by a `<!--more-->` tag, and will use that full excerpt as the `Note`, if present. Otherwise, it uses only the first paragraph as the `Note`.

This could likely be improved, e.g. by being sure to include a minimum amount of text in the `Note`. But if you have any thoughts, please let me know!

---


thebaer:
Per some private discussions, it seems Mastodon will actually show our `Article` preview / excerpt text _today_ by populating the `summary` field on it, so I've also made that change here.

thebaer:
This is good to go now!

PR #1411, opened 2025-07-12
implement postgresql/postgres/pgsql support

leo9800:
WIP

 - `writefreely config start` passed without error
 - `writefreely serve` started without error
 - something on SQL dialect beyond name of datatype needs to be modified, I had encountered (and addressed) some of this type of issues, e.g.
    - `LIMIT x,y` is not supported by postgres, use `LIMIT y OFFSET x` instead
    - `LastInsertId()` is not supported by postgres, use `RETURNING id` with `QueryRow().Scan()` instead
 - not sure if everything works, more intensive testing required

help wanted ...


thebaer:
Thank you so much for tackling this! Would love to get it merged. I'm going to fix the conflicts with `go.mod`, and then leave some comments just from reading through the code. Otherwise, I'm not an expert on Postgres, so I'll ask around to see who might be able to lend some extra help with this.

database.go
```diff
@@ -158,41 +163,76 @@ type datastore struct {
 var _ writestore = &datastore{}
 
 func (db *datastore) now() string {
-	if db.driverName == driverSQLite {
+	switch db.driverName {
+	case driverSQLite:
 		return "strftime('%Y-%m-%d %H:%M:%S','now')"
+	case driverMySQL:
```
thebaer: For switch statements like this, you should be able to combine cases just to keep things as clean as possible, e.g.

```
case driverMySQL, driverPostgres:
```

database.go
```diff
@@ -1321,7 +1374,11 @@ func (db *datastore) GetPosts(cfg *config.Config, c *Collection, page int, inclu
 
 	limitStr := ""
 	if page > 0 {
-		limitStr = fmt.Sprintf(" LIMIT %d, %d", start, pagePosts)
+		if db.driverName == driverPostgres {
+			limitStr = fmt.Sprintf(" LIMIT %d OFFSET %d", pagePosts, start)
+		} else {
+			limitStr = fmt.Sprintf(" LIMIT %d, %d", start, pagePosts)
+		}
```
thebaer: I think we can extract these `LIMIT` clauses into a new helper func that works with any dialect, so we don't have to repeat this logic every time.

db/create.go
```diff
@@ -66,65 +66,80 @@ var UnsetSize OptionalInt = OptionalInt{Set: false, Value: 0}
...
-			if size.Set {
-				mod = fmt.Sprintf("(%d)", size.Value)
-			}
+		switch dialect {
+		case DialectSQLite:
+			return "INTEGER", nil
+		case DialectMySQL:
 			return "INT" + mod, nil
+		case DialectPostgres:
+			return "INT", nil
```
olepbr: PostgreSQL supports `INTEGER`, so you can combine it with the SQLite case here

db/create.go
```diff
@@ -66,65 +66,80 @@ var UnsetSize OptionalInt = OptionalInt{Set: false, Value: 0}
...
 	case ColumnTypeText:
-		return "TEXT", nil
+		switch dialect {
+		case DialectSQLite:
+			return "TEXT", nil
+		case DialectMySQL:
+			return "TEXT", nil
+		case DialectPostgres:
+			return "TEXT", nil
+		}
```
olepbr, line 136: This seems unnecessary; why switch when they're all the same?

db/create.go
```diff
@@ -150,9 +165,15 @@ func (c *Column) SetDefault(value string) *Column {
...
-	if c.Dialect == DialectSQLite {
+	def := ""
+
+	switch c.Dialect {
+	case DialectSQLite:
 		def = "CURRENT_TIMESTAMP"
+	case DialectMySQL:
+		def = "NOW()"
+	case DialectPostgres:
+		def = "NOW()"
```
olepbr: PostgreSQL supports `CURRENT_TIMESTAMP`, so this can be combined with the SQLite case. (MySQL supports `CURRENT_TIMESTAMP`, too, as long as you don't need to support MySQL versions below 5.6, which is 12 years old at this point.)

migrations/drivers.go
```diff
@@ -16,95 +16,238 @@ import (
...
 // TODO: use now() from writefreely pkg
 func (db *datastore) now() string {
-	if db.driverName == driverSQLite {
+	switch db.driverName {
+	case driverSQLite:
 		return "strftime('%Y-%m-%d %H:%M:%S','now')"
+	case driverMySQL:
+		return "NOW()"
+	case driverPostgres:
+		return "NOW()"
```
olepbr: Combine. (Though I think this whole switch could potentially be simplified to `CURRENT_TIMESTAMP` for all drivers.)

migrations/drivers.go
```diff
@@ -16,95 +16,238 @@ import (
...
 
 func (db *datastore) typeInt() string {
-	if db.driverName == driverSQLite {
+	switch db.driverName {
+	case driverSQLite:
 		return "INTEGER"
+	case driverMySQL:
+		return "INT"
+	case driverPostgres:
+		return "INT"
```
olepbr: AFAIK, all support `INTEGER`, no need for a switch.

migrations/drivers.go
```diff
@@ -16,95 +16,238 @@ import (
...
 	}
-	return "INT"
+
+	return "" // placeholder
```
olepbr: Not sure how this placeholder makes any sense?

migrations/drivers.go
```diff
@@ -16,95 +16,238 @@ import (
...
 
 func (db *datastore) typeSmallInt() string {
-	if db.driverName == driverSQLite {
+	switch db.driverName {
+	case driverSQLite:
 		return "INTEGER"
+	case driverMySQL:
+		return "SMALLINT"
+	case driverPostgres:
+		return "SMALLINT"
```
olepbr: Combine.

migrations/drivers.go
```diff
@@ -16,95 +16,238 @@ import (
...
-	if db.driverName == driverSQLite {
+	switch db.driverName {
+	case driverSQLite:
 		// From docs: "In SQLite, a column with type INTEGER PRIMARY KEY is an alias for the ROWID (except in WITHOUT
 		// ROWID tables) which is always a 64-bit signed integer."
 		return "INTEGER PRIMARY KEY"
+	case driverMySQL:
+		return "INT AUTO_INCREMENT PRIMARY KEY"
+	case driverPostgres:
+		return "SERIAL PRIMARY KEY"
```
olepbr: The more modern way to do this is

```suggestion
		return "INTEGER GENERATED BY DEFAULT AS IDENTITY PRIMARY KEY"
```

cf. https://www.depesz.com/2017/04/10/waiting-for-postgresql-10-identity-columns/

migrations/drivers.go
```diff
@@ -16,95 +16,238 @@ import (
...
 
 func (db *datastore) collateMultiByte() string {
-	if db.driverName == driverSQLite {
+	switch db.driverName {
+	case driverSQLite:
+		return ""
+	case driverMySQL:
+		return " COLLATE utf8_bin"
+	case driverPostgres:
 		return ""
```
olepbr: Isn't it easier to change the if to check for MySQL and return `""` otherwise?

migrations/drivers.go
```diff
@@ -16,95 +16,238 @@ import (
...
 }
 
 func (db *datastore) engine() string {
-	if db.driverName == driverSQLite {
+	switch db.driverName {
+	case driverSQLite:
+		return ""
+	case driverMySQL:
+		return " ENGINE = InnoDB"
+	case driverPostgres:
```
olepbr: This too can be a `if db.driverName == driverMySQL` instead

migrations/drivers.go
```diff
@@ -16,95 +16,238 @@ import (
...
 func (db *datastore) after(colName string) string {
-	if db.driverName == driverSQLite {
+	switch db.driverName {
+	case driverSQLite:
 		return ""
+	case driverMySQL:
+		return fmt.Sprintf(" AFTER %s", colName)
+	case driverPostgres:
+		return ""
+	}
```
olepbr, line 165: Just check for MySQL and return `""` otherwise

migrations/drivers.go
```diff
@@ -16,95 +16,238 @@ import (
...
+}
+
+func (db *datastore) boolTrue() string {
+	switch db.driverName {
+	case driverSQLite:
+		return "1"
+	case driverMySQL:
+		return "1"
+	case driverPostgres:
+		return "TRUE"
```
olepbr: All three databases understand `TRUE`, no need for a switch

migrations/drivers.go
```diff
@@ -16,95 +16,238 @@ import (
...
+	switch db.driverName {
+	case driverSQLite:
+		return "0"
+	case driverMySQL:
+		return "0"
+	case driverPostgres:
+		return "FALSE"
+	}
+
+	return "" // placeholder
```
olepbr: As with `TRUE`, all support `FALSE`.

migrations/drivers.go
```diff
@@ -16,95 +16,238 @@ import (
...
+	case driverSQLite, driverMySQL:
+		return fmt.Sprintf(" LIMIT %d, %d", offset, size)
+	case driverPostgres:
+		return fmt.Sprintf(" LIMIT %d OFFSET %d", size, offset)
+	}
+
+	return "" // placeholder
+}
+
+func (db *datastore) QueryWrap(q string) string {
```
olepbr, line 195: This function is a bit impenetrable to me – why does postgres need some form of escaping/quoting here?

app.go
```diff
@@ -959,28 +969,24 @@ func adminInitDatabase(app *App) error {
...
-			log.Info("Creating table ??? (Weird query) No match in: %v", parts)
-		}
-		_, err := app.db.Exec(q)
-		if err != nil {
-			log.Error("%s", err)
+		if _, err := app.db.Exec(q); err != nil {
+			log.Error("Error %s executing:\n%s\n", err, q)
 		} else {
-			log.Info("Created.")
+			log.Info("Successfully executing:\n%s\n", q)
```
olepbr:
```suggestion
			log.Info("Successfully executed:\n%s\n", q)
```

database.go
```diff
@@ -158,41 +163,74 @@ type datastore struct {
...
 
 func (db *datastore) clip(field string, l int) string {
-	if db.driverName == driverSQLite {
+	switch db.driverName {
+	case driverSQLite:
 		return fmt.Sprintf("SUBSTR(%s, 0, %d)", field, l)
+	case driverMySQL:
+		return fmt.Sprintf("LEFT(%s, %d)", field, l)
+	case driverPostgres:
+		return fmt.Sprintf("LEFT(%s, %d)", field, l)
```
olepbr: postgres supports both `LEFT()` and `SUBSTR()`, so combine with either of the above cases

database.go
```diff
@@ -158,41 +163,74 @@ type datastore struct {
...
+	switch db.driverName {
+	case driverSQLite:
 		// NOTE: SQLite UPSERT syntax only works in v3.24.0 (2018-06-04) or later
 		// Leaving this for whenever we can upgrade and include it in our binary
-		cc := strings.Join(indexedCols, ", ")
+		return "ON CONFLICT(" + cc + ") DO UPDATE SET"
+	case driverMySQL:
+		return "ON DUPLICATE KEY UPDATE"
+	case driverPostgres:
 		return "ON CONFLICT(" + cc + ") DO UPDATE SET"
```
olepbr: Combine with SQLite case

database.go
```diff
@@ -158,41 +163,74 @@ type datastore struct {
...
+
+	return "" // placeholder
 }
 
 func (db *datastore) dateAdd(l int, unit string) string {
-	if db.driverName == driverSQLite {
+	switch db.driverName {
+	case driverSQLite:
+		// NOTE: SQLite UPSERT syntax only works in v3.24.0 (2018-06-04) or later
+		// Leaving this for whenever we can upgrade and include it in our binary
```
olepbr: This comment seems unrelated?

database.go
```diff
@@ -158,41 +163,74 @@ type datastore struct {
...
 		return fmt.Sprintf("DATETIME('now', '%d %s')", l, unit)
+	case driverMySQL:
+		return fmt.Sprintf("DATE_ADD(NOW(), INTERVAL %d %s)", l, unit)
+	case driverPostgres:
+		return fmt.Sprintf("DATE_ADD(NOW(), %d %s)", l, unit)
```
olepbr: Not terribly important, but a more idiomatic way would be

```suggestion
		return fmt.Sprintf("NOW() + INTERVAL '%d %s')", l, unit)
```

database.go
```diff
@@ -158,41 +163,74 @@ type datastore struct {
...
+
+	return "" // placeholder
 }
 
 func (db *datastore) dateSub(l int, unit string) string {
-	if db.driverName == driverSQLite {
+	switch db.driverName {
+	case driverSQLite:
+		// NOTE: SQLite UPSERT syntax only works in v3.24.0 (2018-06-04) or later
+		// Leaving this for whenever we can upgrade and include it in our binary
```
olepbr: Again, accidentally copied comment?

database.go
```diff
@@ -158,41 +163,74 @@ type datastore struct {
...
 		return fmt.Sprintf("DATETIME('now', '-%d %s')", l, unit)
+	case driverMySQL:
+		return fmt.Sprintf("DATE_SUB(NOW(), INTERVAL %d %s)", l, unit)
+	case driverPostgres:
+		return fmt.Sprintf("DATE_SUBTRACT(NOW(), %d %s)", l, unit)
```
olepbr:
```suggestion
		return fmt.Sprintf("NOW() - INTERVAL '%d %s')", l, unit)
```

olepbr (commented):
With the caveat that I do not know Go nor this codebase particularly well, here are some comments =)

leo9800:
3 cases for sqlite, mysql and postgres were created in each switch for potential further changes introduced into those sql dialects.

given that the syntax of those sql dialects are not likely to be modified in near future, as it almost certain to be a breaking change and therefore force downstream applications to be reimplemented, I agree that those cases in switches could be merged if possible.

migrations/drivers.go
leo9800: those placeholders are for potential wrong values in `db.driverName` and make the compiler and/or language server happy. (they may not accept an non-`nil`-returning function returning nothing)

if the code execution reaches those placeholders it indicates something goes seriously wrong, (with configuration) maybe we should `panic()` instead of returning an empty string.

migrations/drivers.go
leo9800: sqlite and mysql accepts `?` for placeholder of parameters in prepared queries.

while postgres accepts `$1`, `$2`, etc. this function conducts the convertion, if `db.driverName` is `driverPostgres`.

database.go
leo9800: yep my bad XD

database.go
leo9800: 😅

db/alter.go
```diff
@@ -5,48 +5,69 @@ import (
...
 	return b
 }
 
 func (b *AlterTableSqlBuilder) ChangeColumn(name string, col *Column) *AlterTableSqlBuilder {
 	if colVal, err := col.String(); err == nil {
-		b.Changes = append(b.Changes, fmt.Sprintf("CHANGE COLUMN %s %s", name, colVal))
+		b.Changes = append(b.Changes, change{queryType: alter, queryString: fmt.Sprintf("RENAME COLUMN %s TO %s_old", name, name)})
+		b.Changes = append(b.Changes, change{queryType: alter, queryString: fmt.Sprintf("ADD COLUMN %s", colVal)})
+		b.Changes = append(b.Changes, change{queryType: update, queryString: fmt.Sprintf("SET %s = %s_old", col.Name, name)})
+		b.Changes = append(b.Changes, change{queryType: alter, queryString: fmt.Sprintf("DROP COLUMN %s", name)})
```
olepbr: Shouldn't the last line here be:


```suggestion
		b.Changes = append(b.Changes, change{queryType: alter, queryString: fmt.Sprintf("DROP COLUMN %s_old", name)})
```

Otherwise you end up dropping the newly created column, I think.

I'm basing this on how it's used in `migrations/v5.go`, which, if I'm reading the code correctly, would turn into these SQL statements, given `name = "remote_user_id"` and `col = Column("remote_user_id", wf_db.ColumnTypeVarChar, wf_db.OptionalInt{Set: true, Value: 128})`

```sql
ALTER TABLE oauth_users RENAME COLUMN remote_user_id TO remote_user_id_old;
ALTER TABLE oauth_users ADD COLUMN remote_user_id VARCHAR(128);
UPDATE oauth_users SET remote_user_id = remote_user_id_old;
ALTER TABLE oauth_users DROP COLUMN remote_user_id; -- wrong!
```

~~Overall, I think this function is probably better off as a switch, so that one can keep using `CHANGE COLUMN` for MySQL and use `ALTER COLUMN … SET DATA TYPE …` for PostgreSQL.~~

EDIT: Actually, `ALTER COLUMN … SET DATA TYPE …` won't be enough, since `CHANGE COLUMN` supports specifying a full column definition (e.g. including nullability), which `SET DATA TYPE` does not. So the proposed change (modulo the wrong `DROP COLUMN` statement) is probably the easiest.

migrations/drivers.go
olepbr: Ah, I see, thanks for the insight! I'm no Go expert, but to me it sounds like a panic is more appropriate than generating invalid SQL.

migrations/drivers.go
olepbr: Ah, I see, thanks. I'm kind of surprised that that's something we need to roll ourselves, but maybe the Go postgres driver is quite bare-bones. I think a comment would be apt here.

read.go
```diff
@@ -78,7 +78,7 @@ func (app *App) FetchPublicPosts() (interface{}, error) {
 	FROM collections c
 	LEFT JOIN posts p ON p.collection_id = c.id
 	LEFT JOIN users u ON u.id = p.owner_id
-	WHERE c.privacy = 1 AND (p.created <= ` + app.db.now() + ` AND pinned_position IS NULL) AND u.status = 0
+	WHERE c.privacy = `+app.db.BoolTrue()+` AND (p.created <= ` + app.db.now() + ` AND pinned_position IS NULL) AND u.status = 0
```
olepbr: `collections.privacy` is declared as `INT NOT NULL` IN `postgres.sql`, so this will not work unless that is changed to be a proper boolean. (I suspect there might be other cases that uses integers in place of booleans, since both MySQL and SQLite don't have proper boolean types.)

olepbr (commented):
A few more things I didn't spot last night, will have another look after work :-)

migrations/drivers.go
leo9800: ok i gonna create a panic function for such conditions and invoke them instead of returning the placeholder empty string.

migrations/drivers.go
leo9800: yep most ORMs handle dialect-specific placeholders but lib/pq, which was used to implement this patch, is not the case.

anyway I would comment this function to make it clear.

read.go
thebaer: So the `privacy` column is actually meant to be an integer -- it maps to this value in a `Post` struct:

https://github.com/writefreely/writefreely/blob/66394fba8e96ad2c0b0b9b5717ccde4418a49bd4/posts.go#L109

It's a bit of a legacy thing, but it's meant to support values beyond 0 and 1. So would say we should keep it as an integer.

leo9800: @thebaer got that point.

i'll revert the commit changing the postgres schema of `post.privacy` and  `collection.privacy`.

also gonna modify go code accordingly.

MicaEcho:
The page admin/users is failing to render

```database.go:2934: Failed selecting from users: pq: LIMIT #,# syntax is not supported```

**Comment by leo9800** [comment]
@MicaEcho thanks pinpointing this issue, fixed.

File: postgres.sql [hunk 26]
```diff
@@ -0,0 +1,148 @@
...
+    "script" TEXT DEFAULT NULL,
+    "format" VARCHAR(8) DEFAULT NULL,
+    "privacy" INT NOT NULL,
+    "owner_id" INT NOT NULL,
+    "view_count" INT NOT NULL
+);
+
+
+CREATE TABLE IF NOT EXISTS "posts" (
+    "id" CHAR(16) PRIMARY KEY,
```
WiredPlay, line 68: CHAR data type will pad stored values with extra spaces and urls using the value will start showing e.g. /[id]%20%20%20%20%20%20%20/edit instead /[id]/edit

possible solutions would be using VARCHAR instead
or strings.TrimSpace(res.ID) at posts.go

PR #1418, opened 2025-07-23
Fix config permission 0.15.1

utkarsh-pandey1-psl:
Added changes to tackle the vulnerability CVE-2025-24337


cmd/writefreely/main.go
```diff
@@ -12,19 +12,40 @@ package main
...
 	"github.com/writefreely/writefreely"
 )
 
+
+
 func main() {
 	cli.VersionPrinter = func(c *cli.Context) {
 		fmt.Printf("%s\n", c.App.Version)
 	}
+	cfgPath := "config.ini" // Set your config path logic here
```
thebaer, line 31: We shouldn't hard-code the configuration filename, as it's supplied by the `-c` flag. (See `defaultFlags` in this file below.)

routes.go
```diff
@@ -29,6 +29,17 @@ func (app *App) InitStaticRoutes(r *mux.Router) {
...
+	// 	if strings.HasSuffix(r.URL.Path, "config.ini") {
+	// 		http.NotFound(w, r)
+	// 		return
+	// 	}
+	// 	fs.ServeHTTP(w, r)
+	// })
+
+	// app.shttp = http.NewServeMux()
+	// app.shttp.Handle("/", blockedFS)
+	// r.PathPrefix("/").Handler(blockedFS)
```
thebaer, line 42: It looks like this code isn't meant to be used, so we shouldn't include that here.

thebaer (changes requested):
Thanks for taking a look at this! There are a few changes that need to be made before I can merge this.

thebaer:
Would also appreciate if others could test this out, especially on any non-Linux system!

PR #1459, opened 2025-08-14
Fix Like counts on Stats page

thebaer:
This adds a missing query to populate each post with the number of likes it's received. Previously, this page would always show zero likes.

---


thebaer:
Merging now!

PR #1482, opened 2025-08-26
Use official WriteFreely docs for all help links

thebaer:
This ensures users always go to consistent docs and not our very old site, guides.write.as.

---


thebaer:
Merging now!

PR #1483, opened 2025-08-26
Fix post signature showing up on blog's Post Not Found page

thebaer:
Previously, the signature would show up on Post Not Found pages, as we technically fake a Post behind the scenes to render this page. To fix, this simply inserts our special `<!--nosig-->` shortcode in the body of the post.

---


thebaer:
Works well after testing. Merging now!

PR #1492, opened 2025-09-05
Use datetime picker on Post Metadata page and fix API inconsistency

### LINKED ISSUE #866: Date inconsistencies in the API

thebaer:
This changes the Post Update API to use the RFC3339 format, to fix inconsistencies between format of data returned and taken as input (fixes #866).

In the course of fixing this, we switch to the more user-friendly `datetime-local` form element on the Post Metadata page. This makes it easier for users to choose a date and conforms to the more consisten API.

It finishes the work in #1137.

---

### DIFF
--- a/posts.go
+++ b/posts.go
@@ -48,7 +48,7 @@ const (
 	userPostIDLen = 10
 	postIDLen     = 10
 
-	postMetaDateFormat = "2006-01-02 15:04:05"
+	postMetaDateFormat = "2006-01-02T15:04:05Z"
 )
 
 type PostType string
--- a/templates/edit-meta.tmpl
+++ b/templates/edit-meta.tmpl
@@ -258,7 +258,7 @@
 					<dd><input type="checkbox" id="rtl" name="rtl" {{if .Post.IsRTL.Bool}}checked="checked"{{end}} /><label for="rtl"> right-to-left</label></dd>
 					<dt><label for="created">Created</label></dt>
 					<dd>
-						<input type="text" id="created" name="created" value="{{.Post.UserFacingCreated}}" data-time="{{.Post.Created8601}}" placeholder="YYYY-MM-DD HH:MM:SS" maxlength="19" /> <span id="tz">UTC</span> <a href="#" id="set-now">now</a>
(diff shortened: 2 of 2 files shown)


thebaer:
This will need some more testing, especially with different clients. But we'll merge now.

PR #1505, opened 2025-10-23
Fix collection Customize page header nav on single-user instances

### LINKED ISSUE #1486: Minor 'stats' link bug in 0.16
### Describe the bug
On my single-user instance, the 'stats' link in the "Menu" drop down goes to `/me/c/<myuser>/stats` which works, but the stats link at the top of the "Customise" page is missing the user, so it goes to just `/me/c//stats` which returns a 404.

1. Log in, select 'Menu' then 'stats' -> works
2. Select 'Menu'-> 'customise' -> 'stats' link -> fails with 404

### Expected behavior
Both should behave the same

### Application configuration
 - **Single mode or Multi-user mode?** Single-user mode
 - **Database?**  sqlite
 - **Open registration?**  no
 - **Federation enabled?** yes

**Version or last commit**: 0.16

thebaer:
Previously, the username was blank, messing up the URL of other nav bar links. Likely introduced by SMTP email support changes.

Fixes #1486

---

### DIFF
--- a/templates/user/include/header.tmpl
+++ b/templates/user/include/header.tmpl
@@ -16,9 +16,9 @@
 					</ul>
 				</nav>
 				<nav class="tabs">
-					<a href="/me/c/{{.Username}}" {{if and (hasPrefix .Path "/me/c/") (hasSuffix .Path .Username)}}class="selected"{{end}}>Customize</a>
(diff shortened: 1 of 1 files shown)


thebaer:
Merging now.

PR #1508, opened 2025-10-27
Fix single-user blog URL on Customize page

### LINKED ISSUE #1390: Misleading information in "Customize" section: user name in URL is not working for single-user instance?
Dear maintainers,

I have set up a single-user instance and can reach it, login, create post.

What I found confusing or misleading is the following from the "Customize" section:

```
URL

This blog uses your username in its URL. You can change it in your Account Settings.

    writefreely.example.org/my-username/
```

It is showing a URL with my user name. The problem is that

a) it works when only using `writefreely.example.org`
b) using the URL with the username results in a 404 Not Found.

I guess this could be a missing condition somewhere, that should only add the username to this line, if this is **NOT** a single-user instance?

Kind Regards,
Johannes

thebaer:
Previously, the Customize page showed the wrong URL for single-user blogs, which should always show the top-level address.

Fixes #1390

---


thebaer:
Merging now.

PR #1522, opened 2025-11-18
Redirect /read/feed to correct /read/feed/ URL

thebaer:
The Reader's RSS feed is only served from the URL with the trailing slash.

This fixes an issue originally shared [on the forum](https://discuss.write.as/t/people-kernel-org-redirect-read-feed-to-read-feed/19939).

---

### DIFF
--- a/read.go
+++ b/read.go
@@ -294,7 +294,7 @@ func viewLocalTimelineFeed(app *App, w http.ResponseWriter, req *http.Request) e
 		return impart.HTTPError{http.StatusNotFound, "Page doesn't exist."}
 	}
 	if !strings.HasSuffix(req.URL.Path, "/") {
-		return impart.HTTPError{http.StatusFound, "/read/feed/"}
+		return impart.HTTPError{http.StatusMovedPermanently, "/read/feed/"}
 	}
 
 	updateTimelineCache(app.timeline, false)
--- a/read.go
+++ b/read.go
@@ -17,6 +17,7 @@ import (
 	"math"
 	"net/http"
 	"strconv"
+	"strings"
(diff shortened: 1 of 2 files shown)


read.go
```diff
@@ -292,6 +293,9 @@ func viewLocalTimelineFeed(app *App, w http.ResponseWriter, req *http.Request) e
 	if !app.cfg.App.LocalTimeline {
 		return impart.HTTPError{http.StatusNotFound, "Page doesn't exist."}
 	}
+	if !strings.HasSuffix(req.URL.Path, "/") {
+		return impart.HTTPError{http.StatusFound, "/read/feed/"}
```
matttbe: I'm not a web expert, but would it not be better with a `StatusMovedPermanently` (301) code?
Also, just wondering: is this redirection really needed? Maybe only the new route is enough?

matttbe (commented):
Thank you for this fix!

I just have one question, but feel free to ignore it if it is not relevant!

matttbe (approved):
(I forgot to mark it as "Approve", not sure if that was needed or not, but I'm sure my questions were not blocking ones :) )

read.go
thebaer: Great point -- that *should* help reduce load from RSS readers that respect those types of responses.

As for the redirect itself, it's best for web content to live at a single canonical URL. This is especially true for search engines, but also for RSS readers that might see the same feed (functionally) that lives at two different URLs, and then could end up double-polling your site and adding load to it. So the redirect helps enforce that here.

thebaer:
All good, thank you for the feedback! Merging this now.

PR #1608, opened 2026-02-13
Fix character set for post signature column on MySQL / MariaDB instances

### LINKED ISSUE #1572: Emojis in signature cause a 500 error.
### Describe the bug
When adding emojis to the post signature, a 500 error is raised.

1. Go to customize blog
2. Add 🐱 as the text
3. Press save

### Expected behavior
Signature saved correctly

### Application configuration
- Single Mode
 - database: mysql / mariadb
 - registration: no
 - federation: yes

**Version or last commit**: Latest / v0.16.0

thebaer:
Previously, we never specified the character set, and it seemed to default to `utf8mb3` instead of `utf8mb4`, which caused issues when updating the value if, for example, the signature contained emoji.

Fixes #1572

---

### DIFF
--- a/migrations/migrations.go
+++ b/migrations/migrations.go
@@ -56,22 +56,23 @@ func (m *migration) Migrate(db *datastore) error {
 }
 
 var migrations = []Migration{
-	New("support user invites", supportUserInvites),                 // -> V1 (v0.8.0)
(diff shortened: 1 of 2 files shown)


thebaer:
Merging now.

PR #1630, opened 2026-03-04
Sort list of Fediverse followers by subscription date, descending

thebaer:
This fixes an issue where the Subscribers list could show everyone in an arbitrary order. It also changes formatting of the query to make it a little easier to read.

---


thebaer:
Don't think there's anything else left to do on this -- merging now.

PR #1644, opened 2026-04-02
Fix ProseMirror HTML handling

### LINKED ISSUE #671: Classic editor replaces <pre> with read more
It looks like the classic editor really dislikes anything codeblock. It removes at least linebreaks when codeblock are selected in the dropdown, creates two blocks if two lines already are `<code>` (as wysiwyg at least) and even worse: `<pre>` gets replaced with a readmore insert. After you edit your post. 
At this point i got scared and switched back to the normal markdown editor.

By the way, would be great if the editors were a user option.

thebaer:
Prevent ProseMirror from breaking on unknown HTML
    
Previously, loading any HTML into the Classic editor (such as when switching between editors, mentioned in #671) would cause the entire editor to break or for inline elements to be rendered as "Read more" elements. It simply displays the HTML and maintains it across edits, so no data is lost.

Now, this preserves all HTML loaded into a post, alongside shortcodes like `<!--more-->` and `<!--emailsub-->`.

This **also** fixes errors when building ProseMirror in Node.js 17+.

---


thebaer:
Merging now

PR #1679, opened 2026-06-20
make OAuth Adapter RFC-compliant

### LINKED ISSUE #1241: Generic OIDC breaks RFC6749: Section 2.3
### Describe the bug
I set up Writefreely to use OIDC with Authelia. This does not work without extra steps since i get an error:

`Access Request failed with error: The request is missing a required parameter, includes an invalid parameter value, includes a parameter more than once, or is otherwise malformed. Client Authentication failed with more than one known authentication method included in the request which is not permitted. The registered client with id 'blog.simonszu.de' and the authorization server policy does not permit this malformed request. The `token_endpoint_auth_method` methods determined to be used were 'client_secret_basic', 'client_secret_post'." method=POST path=/api/oidc/token remote_ip=172.19.1.32 stack="github.com/authelia/authelia/v4/internal/handlers/handler_oidc_token.go:25`

1. Configure the oauth.general section to connect Writefreely to Authelia
2. Try to associate the existing user with an Authelia user
3. Get an error in the WebUI, and the above log message in the logs

### Expected behavior
Tokens should be exchanged and the OIDC handshake should succeed

### Application configuration
 - **Single mode or Multi-user mode?** Single
 - **Database?** [mysql/sqlite] MySQL
 - **Federation enabled?** [yes/no] yes

**Version or last commit**: 0.15.1

### Workaround
After some research i found this setting: https://www.authelia.com/configuration/identity-providers/openid-connect/clients/#allow_multiple_auth_methods - which, when enabled, makes everything work. So, in accordance to the Authelia documentation and in regards of the necessity to enable this setting, i raise this issue to double-check the authentication methods which WF signals to the OIDC provider, since apparently Authelia has the strong opinion that WF is currently violating RFC6749 Section 2.3.

mpldr:
RFC 6749 §2.3.1 states: "The client MUST NOT use more than one authentication method in each request." Currently, writefreely always sends both thus requiring a compliant OAuth provider to reject the request. Add an option to send the credentials in the body, if required and disable Basic Auth in this case.

Link: https://datatracker.ietf.org/doc/html/rfc6749#section-2.3
Fixes: #1241

---

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -102,21 +102,22 @@ type (
 	}
 
 	GenericOauthCfg struct {
-		ClientID         string `ini:"client_id"`
-		ClientSecret     string `ini:"client_secret"`
-		Host             string `ini:"host"`
-		DisplayName      string `ini:"display_name"`
-		CallbackProxy    string `ini:"callback_proxy"`
-		CallbackProxyAPI string `ini:"callback_proxy_api"`
-		TokenEndpoint    string `ini:"token_endpoint"`
-		InspectEndpoint  string `ini:"inspect_endpoint"`
-		AuthEndpoint     string `ini:"auth_endpoint"`
-		Scope            string `ini:"scope"`
-		AllowDisconnect  bool   `ini:"allow_disconnect"`
-		MapUserID        string `ini:"map_user_id"`
-		MapUsername      string `ini:"map_username"`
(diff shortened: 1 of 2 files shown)


thebaer:
Thank you for fixing this! Upon reading through, all looks good. Merging now.

PR #1686, opened 2026-07-01
Prevent signups via /auth/signup with closed registrations

### LINKED ISSUE #1649: Registration possible via /auth/signup when open_registration=false
Currently it's possible to register a new user on a multiuser writefreely server by sending a POST request to `/auth/signup`  and despite setting `open_registration=false` in the configuration a fully usable new user is created:

```
curl https://my.writefreely.server/auth/signup \
   -H "Content-Type: application/json" \
   -X POST \
   -d '{"alias": "TestUser", "pass": "SuP4SecuR3", "email": "bla.fas.el"}' 
```

Are you aware and is this intended? Can we prevent it?

PS: not sure if it's actually a security issue, but we tried to reach out to security@writefreely.org beforehand, but haven't received a reply

thebaer:
Previously, there was no check in place on the web-based /auth/signup endpoint to ensure users can't still be created while registrations are closed.

Fixes #1649

---

### DIFF
--- a/unregisteredusers.go
+++ b/unregisteredusers.go
@@ -44,6 +44,9 @@ func handleWebSignup(app *App, w http.ResponseWriter, r *http.Request) error {
 			return ErrBadFormData
 		}
 	}
+	if !app.cfg.App.OpenRegistration && ur.InviteCode == "" {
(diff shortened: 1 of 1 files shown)


thebaer:
Merging now.

PR #1687, opened 2026-07-01
Fix missing user nav on blog Customize page

thebaer:
Previously, user navigation and the _New Post_ button wouldn't show up on the blog Customize page.

Continuing the fixes in #1505, this now ensures the `.Username` is always populated in the navigation bar, so these elements show up.

---


thebaer:
Merging now

Issue #659, opened 2023-03-09
nodeinfo not returning activeMonth statistics


markrprior:
When querying an instance for nodeinfo information none of the active instances include an activeMonth value but some do return data in activeHalfyear.
The Usage function in the nodeinfo module is making a similar SQL call to create the two values:
DATE_SUB(NOW(), INTERVAL 1 MONTH) and DATE_SUB(NOW(), INTERVAL 6 MONTH) but only the latter seems to return useful data. This would suggest that there is no data for the last month but there is data in the range months -2 to -6. It's hard to imagine that all instances have suddenly gone silent (and personaljournal.ca has posts dated yesterday).
As an example if you query https://personaljournal.ca/api/nodeinfo then currently the usage section is
```
'usage' => {
                       'localPosts' => 41675,
                       'users' => {
                                    'total' => 10017,
                                    'activeHalfyear' => 2447
                                  }
                     }
```
As you will note there is no 'activeMonth' field at all which is consistent with omitempty.

thebaer:
Thanks for mentioning this! Forgot to link the issue, but this was fixed by #769, and is available in the latest release!

roughnecks:
Not sure if related, but I'm on master from a couple weeks ago

```
writefreely ERROR: 2026/07/06 22:43:17 nodeinfo.go:99: Failed getting 6-month active user stats: near "6": syntax error                                                                       
                                                                                                                                                                                              
writefreely ERROR: 2026/07/06 22:43:17 nodeinfo.go:110: Failed getting 1-month active user stats: near "1": syntax error 
```

Thanks

thebaer:
@roughnecks What database are you using when this comes up?

roughnecks:
sqlite, the standard writefreely.db

EDIT: I'm running git master because of another issue which is fixed there

thebaer:
Thanks! Replicated and fixed in #1704, which will be in the next release.

PR #1704, opened 2026-07-18
Fix nodeinfo Active User stats for SQLite-backed instances

thebaer:
Previously, the `activeHalfYear` and `activeMonth` attributes would always be `0` on SQLite-backed instances, due to a bad database query.

Fixes another issue brought up in #659.

---


thebaer:
Merging now

PR #1705, opened 2026-07-18
Add mass user management via command-line

thebaer:
This adds a new `users` command for reviewing and moderating users in bulk, especially for use when an instance has experienced a wave of spam, or a ton of unwanted signups while registrations were closed, due to the vulnerability (#1649) in WriteFreely 0.16 and earlier.

This adds these subcommands, with the flags below used to filter only likely unwanted accounts:

- `writefreely users list` -- preview matching users (read-only; includes admins)
- `writefreely users silence` -- silence matched users, with confirmation
- `writefreely users delete` -- delete matched users and their content, with confirmation

Each subcommand accepts these flags, which can be combined as AND conditions:

* `--since` / `--until` (creation date)
* `--no-invite` -- only show users who didn't sign up via invite code
* `--no-oauth` -- only show users who didn't sign up via OAuth
* `--max-posts N` -- only show users with `N` posts or fewer

The general workflow is first to run `writefreely users list [filters]` to confirm the set of users you want to take action on. (**Note**: while the `list` subcommand will include admins if applicable, the mass `silence` and `delete` actions will never include them.)

Next, you'll run `writefreely users silence` or `writefreely users delete` _with the same filter flags_ to take action. This will confirm the list of users, and then confirm you want to take action before running.

### Example

To remove likely malicious accounts that have accrued in recent days on your multi-user instance, you could run the following:

```bash
# Review the accounts first
writefreely users list --since 2026-07-16 --no-invite --no-oauth

# If all are correct, delete them (prompts for confirmation)
writefreely users delete --since 2026-07-16 --no-invite --no-oauth
```

On a single-user instance that has accrued any malicious accounts, you can simply run this (your admin account will always be excluded):

```bash
writefreely users delete
```

---

### DIFF
--- a/app.go
+++ b/app.go
@@ -25,6 +25,7 @@ import (
 	"regexp"
 	"strings"
 	"syscall"
+	"text/tabwriter"
 	"time"
 
 	"github.com/gorilla/mux"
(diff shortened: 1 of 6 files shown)


thebaer:
Merging this now.

Issue #1552, opened 2026-01-02
Unable to send SMTP e-mail on blog subscription


Barbapulpe:
### Describe the bug
I have a blog with a simple name "Le blog". I enable e-mail subscriptions in the "customize blog" page. Then a visitor subscribes at the bottom of the blog, the message says it is successful and he should receive an e-mail with a link to confirm subscription. But this e-mail is never sent, and the following errors are logged by WriteFreely :

`Jan 02 21:08:59 writefreely[2662297]: ERROR: 2026/01/02 21:08:59 email.go:108: Continuing, but error decoding new subscription form request: schema: invalid path "fake_password" (and 1 other error)`
`Jan 02 21:08:59 writefreely[2662297]: ERROR: 2026/01/02 21:08:59 mailer.go:166: Unable to send email to : Mail Error: No recipient specified`
`Jan 02 21:08:59 writefreely[2662297]: 2026/01/02 21:08:59 "POST /api/collections/le-blog/email/subscribe" 302 45.565217ms "Mozilla/5.0 (Android 16; Mobile; rv:146.0) Gecko/146.0 Firefox/146.0"`

The first line of error happened again on a second test but with a long string of 39 seemingly rendom alphanumerical characters instead of "fake password".

The second line of error has also been seen in issue #1545 

Obviously my SMTP paramaters are correct and I am receiving an e-mail when doing a password reset (using Postfix as a SMTP relay, as described in issue #1545 )

Also I see the following lines just after the failed subscription (I would have expected to see 1 email sent), but these repeat every minute.

`Jan 02 21:09:31 writefreely[2662297]: 2026/01/02 21:09:31 [jobs] Fetching email publish jobs...`
`Jan 02 21:09:31 writefreely[2662297]: 2026/01/02 21:09:31 [jobs] Running 0 email publish jobs...`
`Jan 02 21:09:31 writefreely[2662297]: 2026/01/02 21:09:31 [jobs] Done.`

### Steps to reproduce (if necessary)
1. Create a blog named "Test" and allow email subscriptions in "customize blog"
2. Visit the blog and subscribe at the bottom of the blog.
3. You are greeted with a green banner saying you will receive a confirmation email with a link to confirm.
4. The email is never sent, and the errors mentioned above are logged thru `systemctl - writefreely -f`
5. Not receiving the confirmation email with the link, the subscription is in fact never validated.

### Expected behavior
E-mail should have been sent with the confirmation link, and no error in logs.

### Application configuration
[...]

Issue #1545, opened 2025-12-27
Unable to send SMTP email on password reset request


Barbapulpe:
### Describe the bug
When using the "forgot password?" link under the login page for an existing user configured with its email address:
- In the WebUI, the message says that the password reset request was successful: "We sent an email to the address associated with this account." in orange text
- But in fact the email is never sent
- The logs of systemd unit say: `ERROR: 2025/12/27 17:02:22 mailer.go:166: Unable to send email to : Mail Error: No recipient specified`

### Steps to reproduce (if necessary)
1. Create a standard user and populate its email address
2. Logout, and then go to login page
3. Click on the link below "forgot password?"
4. The WebUI answers with "We sent an email to the address associated with this account." in orange text
5. The email is not sent, systemd unit logs say: `ERROR: 2025/12/27 17:02:22 mailer.go:166: Unable to send email to : Mail Error: No recipient specified`

### Expected behavior
Email should be sent, and no error in logs.

### Application configuration
 - **Multi-user mode** 
 - **Database** [mysql]
 - **Open registration** [no]
 - **Federation enabled** [yes]
 - **Email** SMTP

**Version or last commit**:
v0.16.0

Barbapulpe:
Additional info: it seems the error is due to the `account.go` file where email sender addresses are hardcoded as `noreply-password@maildomain.com` (line 1389) and `noreply-login@maildomain.com` (line 1441).

Indeed I get the following logs (edited my email `domain` to `maildomain.com`, `smtp_username` to `smtpuser@maildomain.com` and user name (requesting the password reset) to `xxx@example.net` for anonymity:

`Dec 28 16:21:46 writefreely[93382]: ERROR: 2025/12/28 16:21:46 mailer.go:166: Unable to send email to : Mail Error: No recipient specified`
`Dec 28 16:21:47 writefreely[93382]: ERROR: 2025/12/28 16:21:47 mailer.go:166: Unable to send email to <xxx@example.net>: 553 5.7.1 <noreply-password@maildomain.com>: Sender address rejected: not owned by user smtpuser@maildomain.com`
`Dec 28 16:21:47 writefreely[93382]: ERROR: 2025/12/28 16:21:47 account.go:1369: Error emailing password reset: 553 5.7.1 <noreply-password@maildomain.com>: Sender address rejected: not owned by user smtpuser@maildomain.com`

The sender email should not be hardcoded with such usernames.

[...]

PR #1721, opened 2026-08-07
Fix unquoted From address and phantom recipient in mailer

### LINKED ISSUE #1551: Unable to subscribe to a blog when its name contains commas (or other "special" characters)
### Describe the bug
With a blog names "A, B, C et D", I opened email subscription. The field appears to visitors, but when they try to subscribe to the blog, they get a server error in the Web UI, and the error logs state:

`Jan 01 20:41:56 writefreely[2662297]: ERROR: 2026/01/01 20:41:56 mailer.go:166: Unable to send email to <recipient@example.com>: Mail Error: mail: missing '@' or angle-addr; Header: [From] Address: [A, B, C et D <sender@example.net>]`
`Jan 01 20:41:56 writefreely[2662297]: ERROR: 2026/01/01 20:41:56 email.go:155: Failed to send subscription confirmation email: Mail Error: mail: missing '@' or angle-addr; Header: [From] Address: [A, B, C et D <sender@gayexample.net>]`

I believe this is because the RFC does not allow commas in the e-mail sender name, and WriteFreely gets confused, expecting several recipients when constructing the email.

To solve this, I believe it should be good if WriteFreely puts double-quotes around the blog name (or sanitizes the name).

1. Name a blog "A, B, C et D"
2. In the "Customize blog", tick to allow e-mail subscriptions
3. Visit the blog and subscribe
4. You get a "server error" page in the WebUI and the arror message above is logged
5. As a result, no e-mail is sent, so link cannot be confirmed and subscription is not validated.

### Expected behavior
Subscription should be successful and no error in logs.

### Application configuration
 - **Multi-user mode** 
 - **Database** [mysql]
 - **Open registration** [yes]
 - **Federation enabled** [yes]
 - SMTP e-mail configured

**Version or last commit**:
v0.16.0

thebaer:
Blog names containing commas or other RFC 5322 special characters broke the SMTP "From" header, silently discarding the recipient too and preventing any email from sending. This adds mailer.FormatAddress to properly quote/escape display names.

Also fix mailer.NewMessage pre-allocating a phantom empty-address recipient alongside every single-recipient send, which caused a spurious "No recipient specified" error to be logged on every password reset, login link, and subscription confirmation email. (#1545, #1552)

Fixes #1551

---

### DIFF
--- a/account.go
+++ b/account.go
@@ -1395,7 +1395,7 @@ func emailPasswordReset(app *App, toEmail, token string) error {
 	footerPara := "Didn't request this password reset? Your account is still safe, and you can safely ignore this email."
 
 	plainMsg := fmt.Sprintf("We received a request to reset your password on %s. Please click the following link to continue (or copy and paste it into your browser): %s/reset?t=%s\n\n%s", app.cfg.App.SiteName, app.cfg.App.Host, token, footerPara)
-	m, err := mlr.NewMessage(app.cfg.App.SiteName+" <noreply-password@"+app.cfg.Email.Domain+">", "Reset Your "+app.cfg.App.SiteName+" Password", plainMsg, fmt.Sprintf("<%s>", toEmail))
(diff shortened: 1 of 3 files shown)


thebaer:
Merging now.

PR #1727, opened 2026-08-10
Fix password-protected blogs not unlocking after cookie expires

### LINKED ISSUE #298: Password-protected blog doesn't show correctly
### Describe the bug
I have a password-protected blog which was working as expected a while ago (i.e. tested using different devices and browsers). Today I tried to read the blog from another device, say device A, but I forgot the password. So I logged in on device B and changed the password. But I cannot view the blog from A even when the password is correct. 

I think the authentication actually passed, because wrong inputs will prompt "Incorrect Password" on the page, while the page returned doesn't have this section. 

I tried with another blog and it failed as well. I wonder what might have been broken, because it used to work a month ago. 

Is there any error log I can look into?

1. Switch to passwd-protected mode and set password
2. Try to input password from elsewhere
3. Page still shows input form

### Expected behavior
What should've happened?

Should've been able to read the blog.

### Application configuration
 - **Single mode or Multi-user mode?** Multi-user mode?
 - **Database?** sqlite
 - **Open registration?** no
 - **Federation enabled?** yes
**Version or last commit**: 0.11

thebaer:
Previously, we would only set a new cookie to replace the older, expired one if there was absolutely no error. Instead, we now safely ignore any errors and attempt to re-set the cookie, which should fix this.

Fixes #298

---

### DIFF
--- a/collections.go
+++ b/collections.go
@@ -1409,12 +1409,14 @@ func handleWebCollectionUnlock(app *App, w http.ResponseWriter, r *http.Request)
 
 	// Success; set cookie
 	session, err := app.sessionStore.Get(r, blogPassCookieName)
-	if err == nil {
-		session.Values[readReq.Alias] = true
-		err = session.Save(r, w)
-		if err != nil {
-			log.Error("Didn't save unlocked blog '%s': %v", readReq.Alias, err)
-		}
+	if err != nil {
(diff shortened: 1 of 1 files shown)


thebaer:
Deployed on Write.as now. Merging

PR #1736, opened 2026-08-11
Support serving robots.txt on single-user instances

### LINKED ISSUE #1427: robots.txt not served in single user mode
### Describe the bug

In single user mode `robots.txt` is not served correctly.

1. Drop a `robots.txt` file into `static`
2. `curl http://localhost:8080/robots.txt`

### Expected behavior

I should see contents of `robots.txt`.

Instead I get "Post not found".  The cause seems to be that `viewCollectionPost` doesn't handle `robots.txt`, unlike `handleViewPost`.

I've patched this locally to get `robots.txt` working (see <https://codeberg.org/swsnr/writefreely/commit/993cd4436d026932677876d86841f8dd5bb88044>), but I'm hesitating to make a pull request as I'm not sure this is the right way to do this.  The logic here is quite confusing with this `isRaw`, `handleViewPost` and `viewCollectionPost` split, and I'm not quite so familiar with Go that I could endeavor to refactor this.

### Application configuration
 - **Single mode or Multi-user mode?** Single user
 - **Database?** [mysql/sqlite] sqlite

**Version or last commit**:
0.15.1

thebaer:
Fixes #1427

---

### DIFF
--- a/handle.go
+++ b/handle.go
@@ -467,7 +467,8 @@ func (h *Handler) WebErrors(f handlerFunc, ul UserLevelFunc) http.HandlerFunc {
 }
 
 func (h *Handler) CollectionPostOrStatic(w http.ResponseWriter, r *http.Request) {
-	if strings.Contains(r.URL.Path, ".") && !isRaw(r) {
+	// This mirrors handleViewPost() to serve reserved static files like robots.txt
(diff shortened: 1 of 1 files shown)


thebaer:
Merging now.

PR #1738, opened 2026-08-12
Fix shadowed .Username on collection customize page

snullp:
EmailCfg.Username shadows the promoted StaticPage.Username in the customize struct, so the "blog uses your username in its URL" note never rendered. Use .UserPage.Username.

---


thebaer:
Thanks for fixing this! Looks great, merging now.

PR #1724, opened 2026-08-09
Validate invite codes on /auth/signup with closed registrations

thebaer:
Originally from @snullp in #1722:

> Despite previous fixing efforts, an user can still by created on a closed instance by providing a non-empty invite code to /auth/sighup - nothing validates the code.
>
> Note: the CreateInvitedUser() in signupWithRegistration() is purely cosmetic, it just logs the invite code in DB.

Fixes #1723

---

- [x] I have signed the [CLA](https://todo.musing.studio/L1)

PR #1748, opened 2026-08-20
Enforce signup checks consistently across all registration paths

thebaer:
Previously, registration eligibility (closed registration, invite validity, and password-auth availability) was checked inconsistently depending on the endpoint, including /api/auth/signup, /oauth/signup, and instances with password auth disabled.

Funnels all paths (web, API, OAuth) through a single canRegister() gate so the checks can't be bypassed by picking a different endpoint, and includes the invite code in the OAuth signature. Extends the invite validation from #1724 to the API and OAuth paths it didn't cover. Also adds regression tests covering each path.

---


thebaer:
Merging now.

PR #1763, opened 2026-09-11
Add postgres support

Timshel:
Based on the previous PR (https://github.com/writefreely/writefreely/pull/1411) which appears stalled.

Merged master and added a missing QueryWrap.
Switched all `CHAR/VARCHAR` field to `TEXT` since Postgres has no issue indexing it, and it will prevent issues with length or padding as recently reported in the original PR.

For those wanting to test it: https://github.com/Timshel/writefreely/pkgs/container/writefreely

---


Timshel:
Reverted the modification in `alter.go`, v5 migration was failing with mariadb (https://github.com/writefreely/writefreely/pull/1763/changes/68d1013d9684a8ac43b45b88fe255fe659ce2543).

Changed the v5 migration since contrary to the comment it was changing the type not just the column length.
Was simpler to fix the v4 migration to directly use the correct type, this way new db are ok, and existing one on mysql already ran the past migration. 

This made the `ChangeColumn` unused so removed it since it was only partially implemented.

Timshel:
Tested migrating my instance.
Dumped the db with `mysqldump --skip-ssl --no-create-info`.

Had some issues with binary data, especially with the `email`, ended-up just recreating the user.
For the binary data of the keys wrapped it with: 
```bash
convert_to(E'-----BEGIN PUBLIC KEY----- ... -----END PUBLIC KEY-----\n', 'UTF8'),
```

Some text needed to be escaped: `E'ex \r\n'`

Since column cannot be inserted at a specific position needed to specify the order for the `collections` table :

```sql
INSERT INTO collections (id, alias, title, description, style_sheet, script, post_signature, format, privacy, owner_id, view_count) VALUES
```

And set the `SERIAL` with:
```sql
select setval('users_id_seq', (select max(id)+1 as id from users));
select setval('collections_id_seq', (select max(id)+1 as id from collections));
```

